PR #36, opened 2021-12-29
fix(log trace): download log trace

Kinplemelon:
same as 4.4


kerry-emqx:
More work to be confirmed as basic authentication is not default open to public

PR #39, opened 2021-12-30
fix(advanced sub): sub form

Kinplemelon:
just want to confirm why there are two form components here, and whether there will be any bad things if they are deleted😂


kerry-emqx:
Congrats, You find the secret. it's just a ... BUG

PR #48, opened 2022-01-14
fix(login): remove demo code

ysfscream:


jest.config.js
```diff
@@ -1,6 +1,6 @@
 module.exports = {
```
Kinplemelon, line 1: Have not used jest in the project before

PR #64, opened 2022-02-09
refactor(auth): refactor auth section to ts

ysfscream:


src/views/Auth/Authn.vue
```diff
@@ -60,11 +61,11 @@ export default defineComponent({
   setup() {
     const router = useRouter()
     const { t } = useI18n()
-    const authnList = ref([])
+    const authnList = ref<AuthnItem[]>([])
```
Kinplemelon, line 64: wow

ysfscream: yep

PR #71, opened 2022-02-10
fix(dataManager): fix clear form validate

ysfscream:


src/views/Auth/components/DataManager.vue
```diff
@@ -70,17 +71,19 @@
       :title="isEdit ? $t('Base.edit') : $t('Base.add')"
       width="480px"
       v-model="dialogVisible"
+      @open="handleDialogOpen"
     >
       <el-form ref="recordForm" :model="record" :rules="getRules()" label-position="top">
         <el-form-item prop="user_id" :label="getFiledLabel(field)">
-          <el-input v-model="record.user_id" :disabled="isEdit"></el-input>
+          <el-input size="small" v-model="record.user_id" :disabled="isEdit"></el-input>
```
Kinplemelon, line 78: Can this property be set on the form component?

src/views/Auth/components/DataManager.vue
```diff
@@ -283,8 +294,33 @@ export default defineComponent({
       }
       return fieldMap[field]
     }
+    const handleDialogOpen = () => {
+      recordForm.value?.clearValidate()
+      recordForm.value?.resetFields()
+    }
+    const handleSearch = () => {
+      const page = 1
+      console.log(searchVal)
```
Kinplemelon, line 303: for test?

ysfscream: sorry! We should delete it.

src/views/Auth/components/DataManager.vue
ysfscream: Good suggestion

ysfscream: But the size of the UI components, we need to discuss and keep consistent during the redesign

Kinplemelon: 😲

PR #77, opened 2022-02-16
WIP: Rule Topology 

Kinplemelon:
-  split console and republish node
- add img to node and add some mode config


src/hooks/Rule/useTopology.ts
```diff
@@ -0,0 +1,300 @@
...
+  return iconMap.topic
+}
+
+const getIconFromOutputItem = (output: OutputItem) => {
+  if (typeof output === 'string') {
+    const key = `bridge-${getBridgeTypeFromString(output)}`
+    return iconMap[key]
+  } else {
+    return output.function === RuleOutput.Console ? iconMap.console : iconMap.republish
+  }
```
ysfscream, line 83: The previous if already returns without the else

ysfscream (approved):

PR #80, opened 2022-02-17
WIP: Rule Topology

Kinplemelon:
- move topology to the single menu item
- show bridges in the topology
- tooltip for node in the topology

### DIFF
--- a/src/hooks/Rule/useTopology.ts
+++ b/src/hooks/Rule/useTopology.ts
@@ -5,6 +5,8 @@ import { MQTTBridgeDirection, RuleOutput } from '@/types/enum'
 import iconMap from '@/assets/topologyIcon/index'
 import { BridgeType } from '@/types/enum'
 import { BridgeItem, MQTTOut, OutputItem, RuleItem } from '@/types/rule'
+import useTopologyNodeTooltip from './useTopologyNodeTooltip'
+import { OtherNodeType, RuleInputType, RuleOutputType } from './topologyType'
 
 interface EdgeItem {
   source: string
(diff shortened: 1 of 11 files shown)


src/hooks/Rule/useTopology.ts
```diff
@@ -1,40 +1,13 @@
 import { onMounted, ref, onUnmounted, Ref } from 'vue'
 import { getRules, getBridgeList } from '@/api/ruleengine'
 import G6, { Graph, ModelConfig, IGroup } from '@antv/g6'
```
ysfscream, line 3: I haven't used G6, but I was wondering if it would be possible to help bring the Graph part into a separate file to distinguish the logic of the Graph Register from the logic of the Rule?

ysfscream: Maybe we just need to pass some Rule **Topology logic** or **Data parmas** into a function to generate a complete Topology Graph, but just my thoughts.

ysfscream: Would it also be helpful to split the use of Bridge or Rule?

src/i18n/RuleEngine.js
```diff
@@ -235,4 +235,60 @@ export default {
...
+  failure: {
+    zh: '失败',
+    en: 'Failure',
+  },
+  connected: {
+    zh: '已连接',
+    en: 'Connected',
+  },
+  disconnected: {
+    zh: '未连接',
```
ysfscream, line 291: disconnected 这里准确一点的话应该是 已断开连接 断开连接？

ysfscream (approved):

PR #85, opened 2022-02-21
Some Changes in Bridge List Page

Kinplemelon:
And move topology-related files to separate folders


src/hooks/Rule/topology/bridge/useBridgeTypeValue.ts
```diff
@@ -0,0 +1,51 @@
...
+
+export default () => {
+  const bridgeTypeList = [
+    { value: BridgeType.HTTP, label: 'HTTP' },
+    { value: BridgeType.MQTT, label: 'MQTT' },
+  ]
+
+  const getBridgeLabelByTypeValue = (typeValue: BridgeType) => {
+    return getLabelFromValueInOptionList(typeValue, bridgeTypeList)
+  }
```
ysfscream, line 14:
```js
const getBridgeLabelByTypeValue = (typeValue: BridgeType) =>
	getLabelFromValueInOptionList(typeValue, bridgeTypeList)
```

ysfscream (approved):

src/hooks/Rule/topology/bridge/useBridgeTypeValue.ts
Kinplemelon: I'm trying to avoid unsightly line breaks under formatting rules, so can I write like this?

ysfscream: Not a suggestion that needs attention, nor is it important..

ysfscream: YES...

My advice is to not write a carriage return when there is only one line.

But it's not important

Kinplemelon: got it

PR #89, opened 2022-02-22
Added Some Details to Connector And Some Field Changes

Kinplemelon:


src/common/tools.ts
```diff
@@ -106,3 +106,11 @@ export const getLabelFromValueInOptionList = <T>(
...
   return target?.label || ''
 }
+
+export const commonTimeUnits = [
+  { value: 'ms', label: 'ms' },
+  { value: 's', label: 's' },
+  { value: 'm', label: 'm' },
+  { value: 'h', label: 'h' },
+  { value: 'd', label: 'd' },
+]
```
ysfscream, line 116: Why not this?

```
export const commonTimeUnits = ['ms', 's', 'm', 'h', 'd']
```

ysfscream: I think it should be a problem of packaging in the component?

But I should provide options for both

`Label` and `Value` or `plain string arrays`, mainly to help in writing such data structures

src/i18n/RuleEngine.js
```diff
@@ -327,4 +331,56 @@ export default {
...
+    zh: '留空则自动生成',
+    en: 'Empty will be automatically generated',
+  },
+  testTheConnection: {
+    zh: '测试连接',
+    en: 'Test the connection',
+  },
+  connectionSuccessful: {
+    zh: '连接成功',
+    en: 'Connection is successful',
```
ysfscream, line 384: Connected?

src/views/RuleEngine/components/ConnectorDialog.vue
```diff
@@ -5,152 +5,227 @@
...
-      }
-      submitLoading.value = false
-    }
+    : {},
+)
+const connectorTLS = ref(
+  props.edit && props.modelValue.ssl
+    ? _.cloneDeep(props.modelValue.ssl)
+    : _.cloneDeep(tlsParamsDefault),
+)
```
ysfscream, line 129: Isn't it a bit complicated to include expressions as parameters here?

src/views/RuleEngine/components/ConnectorDialog.vue
```diff
@@ -5,152 +5,227 @@
...
+    isTesting.value = false
+  }
+}
+
+const submitConnector = async (isEdit) => {
+  let res
+  submitLoading.value = true
+  const data = getConnectorData()
+  if (isEdit) {
+    const id = connectorData.value.id
```
ysfscream, line 171: const { id } = connectorData.value

Better? Just a suggestion

ysfscream (approved):

src/i18n/RuleEngine.js
Kinplemelon: When I search in English, 'Connection successful' is more; but my grammar is poor, if you are sure the 'connected successful' is right, I will change it.

src/views/RuleEngine/components/ConnectorDialog.vue
Kinplemelon: yes

src/views/RuleEngine/components/ConnectorDialog.vue
Kinplemelon: Actually, this is historical code.. But when I want to try to modify it, I found a bug in this logic, tks

Kinplemelon: No bug, just need to optimize the logic 😵

src/i18n/RuleEngine.js
ysfscream: No, I meant "Connected". I can confirm that this is a better indication that the connection was successful.

Kinplemelon: This msg is pop up after Test The Connection,  is it right to use 'connected'?

PR #93, opened 2022-02-24
Optimize the Code of Component IoT Form

Kinplemelon:


src/types/rule.d.ts
```diff
@@ -22,6 +22,7 @@ export interface OutputItemObj {
   args?: {
     payload: string
     topic: string
+    qos: number
```
ysfscream, line 25: type: 0 | 1 | 2

src/views/RuleEngine/components/SQLTestDialog.vue
```diff
@@ -0,0 +1,133 @@
...
+  }).catch((e) => {
+    testParams.value.output = e
+  })
+
+  if (res) {
+    try {
+      const text = JSON.stringify(res)
+      testParams.value.output = text
+    } catch (e) {
+      console.log(e)
```
ysfscream, line 117: message.error(e)?

ysfscream: If the json content is incorrect, the user should be able to see the error.

ysfscream (approved):

src/types/rule.d.ts
Kinplemelon: 👌🏻

src/views/RuleEngine/components/SQLTestDialog.vue
Kinplemelon: Here is the previous code, I did not modify the logic; I think if the response is not in JSON format?

PR #98, opened 2022-02-27
Monaco Component Support Model Value Binding

Kinplemelon:
- extract the rule outputs component
- updated the props of monaco & fixed a bug

### DIFF
--- a/src/components/Monaco.vue
+++ b/src/components/Monaco.vue
@@ -3,100 +3,108 @@
 </template>
 
 <script>
-import * as monaco from 'monaco-editor'
-import { computed, onMounted, onUnmounted, ref, watch } from 'vue'
-// import * as monaco from "monaco-editor/esm/vs/editor/editor.api";
-// import { createMonacoComplete, createMonacoHover } from "@/common/monacoUtils";
+import { defineComponent } from 'vue'
 
-export default {
+export default defineComponent({
   name: 'Monaco',
+})
+</script>
 
-  props: {
-    id: {
-      type: String,
-      required: true,
-    },
-    value: {
-      type: String,
-      required: true,
-    },
-    lang: {
-      type: String,
-      required: true,
-    },
-    disabled: {
-      type: Boolean,
-      default: false,
-    },
-    // warp: {
-    //   type: Boolean,
-    //   default: false,
-    // },
-    // provider: {
-    //   type: Array,
-    //   default: () => [],
-    // },
-    scrollLoading: {
-      type: Boolean,
-      default: false,
-    },
-    scrollFunc: {
-      type: Function,
-      default: () => () => {},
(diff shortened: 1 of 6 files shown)


src/components/Monaco.vue
```diff
@@ -3,100 +3,108 @@
...
+    default: false,
+  },
+  // warp: {
+  //   type: Boolean,
+  //   default: false,
+  // },
+  // provider: {
+  //   type: Array,
+  //   default: () => [],
+  // },
```
ysfscream, line 41: Should we need these options? I think we should.

src/components/Monaco.vue
```diff
@@ -3,100 +3,108 @@
...
-      }
+const emit = defineEmits(['update:modelValue'])
 
-      editor.value = monaco.editor.create(document.getElementById(id), defaultOptions)
+// ❗️ editor instance can not be reactive, otherwise it will cause the page to get stuck for unknown reasons
+let editor = {}
+// const editValue = computed(() => {
+//   console.log(prop.modelValue + "changed");
+//   return prop.modelValue;
+// });
```
ysfscream, line 59: OK, It shouldn't matter if you don't use reactive here, what does the pseudo-code comment mean?

src/i18n/RuleEngine.js
```diff
@@ -355,6 +359,14 @@ export default {
     zh: 'SQL 语句',
     en: 'SQL',
   },
+  ruleSQLDesc: {
+    zh: '使用 SQL 语句实时提取、过滤、丰富和转换设备与业务系统之间的 IoT 数据。',
+    en: 'Extract, filter, enrich, and transform IoT data between devices and business systems in real-time using SQL statements.',
+  },
+  doc: {
+    zh: '参考文档',
+    en: 'Doc',
```
ysfscream, line 368: Document?

ysfscream (approved):

src/i18n/RuleEngine.js
Kinplemelon: ok

src/components/Monaco.vue
Kinplemelon: It's the previous code, I always don't want to delete the previously commented code, what if I need it later?

src/components/Monaco.vue
Kinplemelon: I haven't figured out the usage of these two yet, I'll add them later.

src/components/Monaco.vue
ysfscream: It looks like for debugging... Ok, just remove it?

ysfscream: BTW, I think the init data of the `editor` will be `null`?

PR #103, opened 2022-03-01
feat(rule): copy and use SQL in rule template drawer

Kinplemelon:


src/views/RuleEngine/components/SQLTemplateDrawer.vue
```diff
@@ -99,6 +110,20 @@ const showDrawer: WritableComputedRef<boolean> = computed({
     emit('update:modelValue', val)
   },
 })
+
+const useSQL = (SQLContent: string) => {
+  emit('use-sql', SQLContent)
+  showDrawer.value = false
+}
+
+const copyText = async (text: string) => {
```
ysfscream, line 119: There is already a method in useCopy hooks that I feel like I can call directly:

https://github.com/emqx/emqx-dashboard-web-new/blob/dev/0.22.0/src/hooks/useCopy.js

ysfscream: Usage ref: https://github.com/emqx/emqx-dashboard-web-new/blob/dev/0.22.0/src/views/Auth/components/DatabaseConfig.vue#L279

Kinplemelon: Some of them are no longer usable, and I will delete the useless code. There is indeed available code in the project, but it doesn't feel good to use. I will unify all copy operations to the new method.

Kinplemelon: You are very familiar with code.

ysfscream: > Some of them are no longer usable, and I will delete the useless code. There is indeed available code in the project, but it doesn't feel good to use. I will unify all copy operations to the new method.

That's great!

ysfscream: There are a few places where you need to check for old code that was used before.

ysfscream (approved):

PR #104, opened 2022-03-01
feat(config): add basic config page

ysfscream:
@Kinplemelon Note:

1. scheam.json is the default config schema form render file, the emqx build will replace it with the new file.
2. Currently in this PR it's just a matter of building the schema.json content into a data format that works well on the front end and printing it to the page

image

### DIFF
--- a/src/hooks/Config/useSchemaForm.ts
+++ b/src/hooks/Config/useSchemaForm.ts
@@ -1,0 +1,92 @@
+import { inject, ref, Ref, watch } from 'vue'
+import _ from 'lodash'
+
+type InjectSchema = Ref<any> | undefined
+
+interface Properties {
+  [key: string]: {
+    $ref?: string
+    description: string
+    label: string
+    default: any
+    symbols?: string[]
+    type: string
+    parent?: string
+    properties?: Properties
+  }
+}
+
+interface Component {
+  type: string
+  properties: Properties
+}
+
+interface Components {
+  schemas: {
+    [key: string]: {
+      type: string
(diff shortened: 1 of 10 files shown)


src/hooks/Config/useSchemaForm.ts
```diff
@@ -0,0 +1,92 @@
...
+          }
+          res[key] = property
+        })
+      }
+      return res
+    }
+    const { $ref } = data.paths[path].get
+    const component = getComponentByRef(data, $ref)
+    const components = transComponents(component)
+    return components
```
ysfscream, line 83: If you think there is a better code, please recommend it to me

src/views/Config/BasicConfig/BasicConfig.vue
```diff
@@ -0,0 +1,52 @@
...
+    Cluster,
+  },
+  setup() {
+    const panes = ref(['cluster'])
+    const { handleClickTab, setPaneRef } = useSubTabMenu(panes.value)
+    const schemaRequest = axios.create({
+      baseURL: '',
+    })
+    const schema = ref({})
+    provide('schema', schema)
```
ysfscream, line 34: Use provide/inject to share the state of the child components as this is part of the configuration page and don't want to use Vuex or pass props many times.

src/hooks/Config/useSchemaForm.ts
Kinplemelon: have no idea 😵, I think it's good

PR #109, opened 2022-03-03
refactor(inputWithUnit): support string arrary type

ysfscream:
@Kinplemelon Input with unit component support pass string array or object array.

It's easier to use, see the code, but I'm not sure if there will be problems, maybe I'll merge first but please help review!!


Kinplemelon:
> @Kinplemelon Input with unit component support pass string array or object array.
> 
> It's easier to use, see the code, but I'm not sure if there will be problems, maybe I'll merge first but please help review!!

Looks good

PR #125, opened 2022-03-09
feat(config): add dashboard config page

ysfscream:
WIP


ysfscream:
Using table component to edit arrays

image

PR #126, opened 2022-03-09
Some Details

Kinplemelon:


src/views/RuleEngine/Bridge/BridgeDetail.vue
```diff
@@ -141,6 +128,11 @@ const loadBridgeInfo = async () => {
   infoLoading.value = false
 }
 
+const goDoc = () => {
+  // TODO:
+  window.open('https://www.emqx.io/', '_blank')
```
ysfscream, line 133: Although it is a TODO status, jumping to an external link is still recommended:
```js
const windowUrl = window.open('https://www.emqx.io/', '_blank')
windowUrl.opener = null;
```

ysfscream (approved):

src/views/RuleEngine/Bridge/BridgeDetail.vue
Kinplemelon: Can I ask why?

ysfscream: When a newly opened page points to a site we don't know about, the new page gains access to some part of the linked page through the `window.opener` object. Although we won't here, it's just a customary way of writing security protection, it's not necessary here, and maybe you could also learn some security issues related to newly opened pages.

PR #134, opened 2022-03-11
feat(config): completed the zone config

ysfscream:


src/components/SchemaForm.tsx
```diff
@@ -28,7 +28,6 @@ const SchemaForm = defineComponent({
       required: true,
     },
     form: {
-      type: Object,
```
Kinplemelon, line 31: why deleted it?

PR #159, opened 2022-03-22
Show Bridge Name When Node Is Bridge & Adjusted  MQTT Out Form Content

Kinplemelon:


src/hooks/useFormRules.ts
```diff
@@ -15,33 +27,45 @@ export default () => {
       },
     ]
   }
-  const createIntFieldRule = (min?: number, max?: number) => {
-    const ret = [
+
+  const createIntFieldRule = (min?: number, max?: number): Array<RuleItem> => {
```
ysfscream, line 31: Is this function somewhat similar to the function here in tools.ts:10? Or can I just use one? There are many places where we need to use this check rule

https://github.com/emqx/emqx-dashboard-web-new/pull/159/files#diff-35169bc4bc0fb8270e4855086040dee50350355e7d87fd14f68422cdd36404d0R10

ysfscream: Oh, I got it. Noting.

ysfscream: But this function could be made a little more generic, and I'm sure there will be many places where it can be used

ysfscream (approved):

src/hooks/useFormRules.ts
Kinplemelon: the func in tools.ts is originally prepared for this place, but latter I found can config rule object to complete the function. 
but the func is used in other place so i did not delete it.
you are turely familiar with code 🤣 
i will think how make it more generic

PR #162, opened 2022-03-23
fix(overview): remove useless codes

ysfscream:
@zmstone It's now called a Websocket Client

@Kinplemelon In the project, as long as the code is not used, currently need to delete, it without comments, because it will bring a lot of interference, there are similar functions we need to re-write the code to reach the latest.

### DIFF
--- a/src/i18n/Auth.js
+++ b/src/i18n/Auth.js
@@ -119,14 +119,6 @@ export default {
     zh: '下移',
     en: 'Move down',
   },
-  enable: {
-    zh: '启用',
-    en: 'Enable',
-  },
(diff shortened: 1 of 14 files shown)


ysfscream:
image

@Kinplemelon I have refactored the interaction of the Auth list in the same way as IoT Rule. But about your table dropdown menu in IoT Rule, I would like to write the two outer buttons to the dropdown component as well, see the implementation in PR.

And then add icons for the buttons in the drop-down menu.

image

Kinplemelon (approved):

Kinplemelon (approved):

PR #166, opened 2022-03-24
feat(auth): add topic tips

ysfscream:
@Kinplemelon tooltip component doesn't seem to have the width props.

### DIFF
--- a/src/i18n/Auth.js
+++ b/src/i18n/Auth.js
@@ -303,4 +303,8 @@ export default {
     zh: '写模式',
     en: 'Write Mode',
   },
+  topicTips: {
+    zh: '主题字符串，支持 ${0} 和 ${1} 占位符，在执行权限检查时会被替换为对应客户端的 Username 或 Client ID',
+    en: "Topic supports ${0} and ${1} placeholders, which will be replaced with the corresponding client's Username or Client ID when performing permission checks.",
+  },
 }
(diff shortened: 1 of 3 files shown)


src/i18n/Auth.js
```diff
@@ -303,4 +303,8 @@ export default {
     zh: '写模式',
     en: 'Write Mode',
   },
+  topicTips: {
+    zh: '主题字符串，支持 ${0} 和 ${1} 占位符，在执行权限检查时会被替换为对应客户端的 Username 或 Client ID',
```
Kinplemelon, line 307: can use "主题字符串，支持 {'{'}username{'}'} 和 {'{'}clientid{'}'} 占位符..."?
like this https://vue-i18n.intlify.dev/guide/essentials/syntax.html#literal-interpolation

Kinplemelon: > @Kinplemelon tooltip component doesn't seem to have the width props.

oh, should I add width prop?

ysfscream: Oh, thanks so much. I've been looking all morning and haven't found out how to fix it. 😂

ysfscream: > > @Kinplemelon tooltip component doesn't seem to have the width props.
> 
> oh, should I add width prop?

YES, it's useful.

PR #187, opened 2022-04-02
fix(plugins): typo in plugin package format description text

zmstone:


src/i18n/Plugins.js
```diff
@@ -84,8 +84,8 @@ export default {
     en: 'Upload file',
   },
   uploadTip: {
-    zh: '如果您有 .eip 格式的插件文件，可以在这里通过上传安装它。您也可以将插件包放置到 data/plugins/ 目录随后到插件列表手动进行安装。',
-    en: 'If you have a plugin file in .eip format, you can upload and install it here. You can also put the plugin package in the data/plugins/ directory and then go to the plug-in list to install it manually.',
+    zh: '您可以在这里上传安装 .tar.gz 格式的 plugin 安装包，或用其他方法将插件包上传到 EMQX 所在服务器的 {data_data}/plugins/ 目录。随后到插件列表进行安装。',
+    en: 'You can upload and install a plugin .tar.gz package from here, or upload the package to {data_dir}/plugins/, then you should be able to see it in the list and install from there.',
```
ysfscream, line 88: Hi, thanks for your help. But  I don't recommend using a syntax like `{ data_data }` here. If we use it this way, we need to use named formatting to change the content of the `data_data`, just like below:

```js
const messages = {
  en: {
    message: {
      hello: '{msg} world'
    }
  }
}
```
Template:
```html
<p>{{ $t('message.hello', { msg: 'hello' }) }}</p>
```
Output:
```html
<p>hello world</p>
```

ysfscream: And could you please modify it and commit it to the https://github.com/emqx/emqx-dashboard-web-new/tree/dev/0.28.0 branch?

ysfscream (changes requested):
And could you please modify it and commit it to the [dev/0.28.0](https://github.com/emqx/emqx-dashboard-web-new/tree/dev/0.28.0) branch?

src/i18n/Plugins.js
ysfscream: Like this should be fine

```suggestion
    zh: '您可以在这里上传安装 .tar.gz 格式的 plugin 安装包，或用其他方法将插件包上传到 EMQX 所在服务器的 data_data/plugins/ 目录。随后到插件列表进行安装。',
    en: 'You can upload and install a plugin .tar.gz package from here, or upload the package to data_dir/plugins/, then you should be able to see it in the list and install from there.',
```

ysfscream (approved):

ysfscream:
LGTM

PR #209, opened 2022-04-14
Some Changes in Gateway & Do Not Effect Progress When Request Current Metrics Data

Kinplemelon:


src/api/common.ts
```diff
@@ -27,7 +27,7 @@ export function loadMetrics(): Promise<Array<NodeStatisticalData>> {
 }
 
 export function loadCurrentMetrics(): Promise<Record<string, number>> {
-  return http.get('/monitor_current')
+  return http.get('/monitor_current', { doNotEffectProgress: true })
```
ysfscream, line 30: `doNotEffectProgress` I think is not a good variable name 😂

Forgive me for saying that.

```suggestion
  return http.get('/monitor_current', { hideTopProgress: true })
```

src/common/http.js
```diff
@@ -30,25 +30,31 @@ axios.interceptors.request.use(
...
-  if (store.state.request_queue) {
-    //nothing
-  } else {
-    NProgress.start()
+  if (!config.doNotEffectProgress) {
+    if (store.state.request_queue) {
+      //nothing
+    } else {
+      NProgress.start()
+    }
```
ysfscream, line 38: Much better than `nothing` comment?
```suggestion
    if (!store.state.request_queue) {
      NProgress.start()
    }
```

ysfscream (approved):

src/api/common.ts
Kinplemelon: but the progress bar is controlled by a request stack, send a request will trigger stack push, a request received response will trigger stack pop, the progress bar will be in progress when the stack is not empty, so i think the 'hideTopProgress'  maybe can't describe the handler precisely

src/common/http.js
Kinplemelon: yep

src/api/common.ts
ysfscream: I just think this parameter to describe the action, the action is `hide`, instead of using it to express technical details, maybe I just don't understand the word `effect`. Hmmm.. so, do what you feel is right.

Kinplemelon: even if the quest stack isn't empty when requesting current metrics, will not hide the progress bar, just not trigger stack push and don't trigger stack pop after received metrics ( i think about the variable name long time, don't let me change it, please QAQ

Kinplemelon: how about 'doNotTouchProgress' :P

ysfscream: Touch -> Trigger?

Kinplemelon: use it!

ysfscream: OK! 😂 Whatever, keep it. It's not important.

PR #211, opened 2022-04-14
Extract Listener Dialog

Kinplemelon:


src/api/listener.ts
```diff
@@ -0,0 +1,18 @@
...
+  return http.post(`/listeners/${id}`, data)
+}
+
+export const updateListener = (data: Listener, id: string): Promise<Listener> => {
+  return http.put(`/listeners/${id}`, data)
+}
+
+export const deleteListener = (id: string): Promise<any> => {
+  return http.delete(`/listeners/${id}`)
+}
```
ysfscream, line 18: In the future, we can add an encodeURIComponent method to our `id` or `username` parameters to prevent problems caused by special characters

ysfscream (approved):

PR #219, opened 2022-04-19
refactor(clients): adapt the client page to gateway

ysfscream:


src/components/DetailHeader.vue
```diff
@@ -28,7 +29,11 @@ const props = defineProps({
 })
 
 const goBack = () => {
-  router.push({ path: props.item.path || '/' })
+  if (props.item.backFunc) {
+    props.item.backFunc()
+  } else {
+    router.push({ path: props.item.path || '/' })
```
Kinplemelon, line 35: maybe can use
```javascript
props.item.path ? router.push({ path: props.item.path }) : router.back()
```

ysfscream: YES. Looks better.

PR #240, opened 2022-04-25
Fixed Some Bugs

Kinplemelon:


src/components/TextareaWithUploader.vue
```diff
@@ -119,8 +119,8 @@ const handleError = (error: any) => {
   .el-button {
     font-weight: normal;
     font-size: 14px;
-    background-color: #f7f7f7;
-    border-color: #f7f7f7;
+    background-color: --color-bg-primary;
+    border-color: --color-border-primary;
```
ysfscream, line 123: Don't we need to use the `var` function here??

```css
background-color: var(--color-bg-primary);
```

Kinplemelon: oh yes!

Kinplemelon: fixed it, PTAL

ysfscream: cool

ysfscream (approved):

PR #244, opened 2022-04-27
Fixed Some Bugs in Rule

Kinplemelon:


src/hooks/Auth/useAuthn.ts
```diff
@@ -40,7 +40,7 @@ export default (): {
   }
 
   const getAuthnList = async (isInit = false) => {
-    isListLoading.value = !isInit && true
+    !isInit ? (isListLoading.value = true) : void 0
```
ysfscream, line 43: Hmmm... You should actually just use an if statement? 😳

```suggestion
   if (!isInit) isListLoading.value = true
```

src/hooks/Auth/useCSSVariables.ts
```diff
@@ -0,0 +1,22 @@
+export default (): {
```
ysfscream, line 1: This file should not be under the Auth folder.

Kinplemelon: oh!

ysfscream (approved):

PR #245, opened 2022-04-27
refactor(topic_metrics): adapt the dark mode to topic metrics

ysfscream:


src/views/Diagnose/TopicMetrics.vue
```diff
@@ -85,7 +85,7 @@
         :label="tl('msgIn')"
         sortable
         :sort-by="({ metrics }) => metrics['messages.in.count']"
-        :min-width="136"
+        :min-width="180"
```
Kinplemelon, line 88: why change it QAQ

PR #258, opened 2022-05-06
fix(i18n): rename setting to settings

ysfscream:


Kinplemelon:
🤣 maybe we should extract some common i18n

Kinplemelon:
I also write some duplicate i18n 🥲

ysfscream:
> 🤣 maybe we should extract some common i18n

YES!! But it just feels like too much there's no time to do it 😭

PR #369, opened 2022-06-21
fix(index): icon in tab

Kinplemelon:
In short, I compared the 4.* version, confirmed the back-end code, and found that the icon can be displayed by changing the path. But I still don't understand why the old path sometimes works and sometimes it doesn't.


Kinplemelon:
I got the card back from the back end

ysfscream (approved):

ysfscream:
> I got the card back from the back end

Thanks!

PR #373, opened 2022-06-29
refactor(websocket): when the connection fails, display the reason for the connection failure

Kinplemelon:


Kinplemelon:
I noticed that the MQTTX web version also does not show the error reason when the connection fails, if you think this PR is no big problem, I will pull a similar PR to MQTTX

As for why the MQTTX client can display errors, I think it is probably due to the different protocols, but because MQTTX has not been able to run successfully on my computer for the past these days, I did not research it detailedly.

ysfscream (approved):

ysfscream:
👍👍👍

ysfscream:
> I noticed that the MQTTX web version also does not show the error reason when the connection fails, if you think this PR is no big problem, I will pull a similar PR to MQTTX
> 
> As for why the MQTTX client can display errors, I think it is probably due to the different protocols, but because MQTTX has not been able to run successfully on my computer for the past these days, I did not research it detailedly.

Thanks. If it's best, you can make a change to MQTTX and MQTT X Web as well!

src/i18n/MQTTRes.js
```diff
@@ -0,0 +1,201 @@
+export default {
+  /* v3{code} */
+  v300: {
+    zh: '已接受连接',
+    en: 'Connection accepted',
+  },
+  v301: {
+    zh: '服务端不支持客户端请求的MQTT协议级别',
```
ysfscream, line 8:
```
服务端不支持客户端请求的 MQTT 协议级别
```
 Need blank

PR #404, opened 2022-07-26
feat(user): force a password change if the user is using the default password

Kinplemelon:


Kinplemelon:
image

Kinplemelon:
to be honest, this can't prevent some user who with frontend knowledge to delete element in the console, but I think it is enough (isn't it?

src/i18n/Base.js
```diff
@@ -435,4 +435,8 @@ export default {
     zh: '主题',
     en: 'Topic',
   },
+  defaultPwdTip: {
+    zh: '您的密码为默认密码，请立即修改',
+    en: 'Your password is the default, please change it immediately',
```
ysfscream, line 440:
```suggestion
    en: 'Your password is the default, please change it immediately',
    en: 'Detected that you are logging in with the default password, please change your password in time. ',
```

src/i18n/Base.js
```diff
@@ -435,4 +435,8 @@ export default {
     zh: '主题',
     en: 'Topic',
   },
+  defaultPwdTip: {
+    zh: '您的密码为默认密码，请立即修改',
```
ysfscream, line 439:
```suggestion
    zh: '您的密码为默认密码，请立即修改',
    zh: '检测到您正在使用默认密码登录，请及时修改密码',
```

src/hooks/useChangePwdGuide.ts
```diff
@@ -0,0 +1,92 @@
...
+  const store = useStore()
+  const router = useRouter()
+  const { tl } = useI18nTl('Base')
+
+  const isUsingDefaultPwd = computed(() => {
+    return store.state.user.isUsingDefaultPwd
+  })
+
+  const popupMessageBox = () => {
+    ElMessageBox.alert(tl('defaultPwdTip'), {
```
ysfscream, line 18: type: info?

ysfscream:
> to be honest, this can't prevent some user who with frontend knowledge to delete element in the console, but I think it is enough (isn't it?

YES. Do not care how others will technically hack, in fact, modify the default password just to make the user more secure.

src/hooks/useChangePwdGuide.ts
Kinplemelon: There is no info type, although it is an alert type but there is no red UI

PR #418, opened 2022-08-04
style(iconfont): diff enterprise and open source

Kinplemelon:


Kinplemelon:
maybe we need to update the build_packages?

ysfscream:
> maybe we need to update the build_packages?

YES.

ysfscream (approved):

PR #431, opened 2022-08-09
refactor(schema form): handle SSL data; submit data when create bridge; do not show tab when type is bridge

Kinplemelon:


src/common/tools.ts
```diff
@@ -371,3 +371,6 @@ export const getLocalMessage = (path: string) => {
   const messages = i18n.global.messages[local]
   return get(messages, path)
 }
+
+export const getBridgeIconKey = (value: string) =>
+  value.indexOf('influxdb') > -1 ? `influxdb` : value
```
ysfscream, line 376: Why the special handling of influxdb icon?

Kinplemelon: image

Kinplemelon: well, there is some mistakes cause CI failed, I see see

ysfscream (approved):

PR #451, opened 2022-08-15
fix(bridge): catch the error that can not load bridge icons

Kinplemelon:


src/hooks/Rule/bridge/useBridgeTypeValue.ts
```diff
@@ -88,3 +88,25 @@ export const useBridgeTypeOptions = (): {
     getTypeStr,
   }
 }
+
+export const useBridgeTypeIcon = (): {
+  getBridgeIconKey: (value: string) => string
+  getBridgeIcon: (type: string) => string
+} => {
+  const getBridgeIconKey = (value: string) => (value.indexOf('influxdb') > -1 ? `influxdb` : value)
```
ysfscream, line 96: influxdb's should be the enterprise version of the code

ysfscream (approved):

PR #509, opened 2022-09-08
refactor(global): replace new-password with one-time-code on autocomp…

Kinplemelon:
…lete prop in password input


Kinplemelon:
use one-time-code will not let the browser popup to ask users "replace new password?"

Kinplemelon:
use one-time-code will not let the browser popup to ask users "replace new password?"

ysfscream (approved):

PR #521, opened 2022-09-14
Line Protocol Form For InfluxDB

Kinplemelon:


src/hooks/Rule/bridge/useInfluxdbLineProtocol.ts
```diff
@@ -0,0 +1,99 @@
...
+  // TODO:Consider the case of spaces or commas inside string value
+  const tagItemReg = /(?<tagKey>[^\s,]+)=(?<tagValue>[^\s,]+)/
+  const tagPartReg = new RegExp(`(${tagItemReg.source},?)+`)
+  const fieldItemReg = /(?<fieldKey>[^\s,]+)((?<!\/)=)(?<fieldValue>[^\s,]+)/
+  const fieldsPartReg = new RegExp(`(${fieldItemReg.source},?)+`)
+  const timestampPart = /\d+/
+
+  const protocolReg = new RegExp(
+    `^(?<measurement>${measurementReg.source})(,(?<tags>${tagPartReg.source}))?(\\s(?<fields>${fieldsPartReg.source}))(\\s(?<timestamp>${timestampPart.source}))?$`,
+  )
```
ysfscream, line 41: I remember there is a place to manage Reg expression constants: https://github.com/emqx/emqx-dashboard-web-new/blob/master/src/common/constants.ts#L49

But there is nothing wrong with writing it here.

ysfscream (approved):

src/hooks/Rule/bridge/useInfluxdbLineProtocol.ts
Kinplemelon: because ip reg can be used in many places but these regs just can be used in this place

PR #524, opened 2022-09-15
fix(bridge): reorder bridge list column & update codes

ysfscream:


src/hooks/Rule/topology/useTopologyBridgeData.ts
```diff
@@ -1,9 +1,8 @@
 import { getBridgeList as queryBridgeList } from '@/api/ruleengine'
-import { BridgeItem, MQTTOut } from '@/types/rule'
+import { BridgeItem } from '@/types/rule'
```
Kinplemelon, line 2: useless import cleaning master ( ok i will try do it too

PR #534, opened 2022-09-19
fix(bridge): pop up warning before update bridge

Kinplemelon:


src/i18n/RuleEngine.js
```diff
@@ -7,6 +7,10 @@ export default {
     zh: '编辑数据桥接',
     en: 'Edit Data Bridge',
   },
+  updateBridgeTip: {
+    zh: '更新后该数据桥接将重新启动，可能造成数据丢失或服务中断，是否确认更新？',
+    en: 'This data bridge will be restarted after the update, which may cause data loss or service interruption, is the update confirmed?',
```
ysfscream, line 12:
```suggestion
    en: 'This data bridge will be restarted after the update, which may cause data loss or service interruption, is the update confirmed?',
    en: 'This data bridge will be restarted after the update, which may cause data loss or service interruption, confirm?',
```

ysfscream (approved):

PR #584, opened 2022-10-19
style(login): change some style for beautify style when auto filled

Kinplemelon:
image

### DIFF
--- a/src/i18n/components.js
+++ b/src/i18n/components.js
@@ -333,7 +333,7 @@ export default {
   },
   integration: {
     zh: '监控集成',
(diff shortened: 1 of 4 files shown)


src/i18n/components.js
```diff
@@ -332,8 +332,8 @@ export default {
     en: 'Retainer',
   },
   integration: {
-    zh: '监控集成',
```
ysfscream, line 335: Why changes here? Please do not modify the Chinese word "监控集成"

ysfscream (approved):

PR #666, opened 2022-11-22
fix(websocket): add session expiry interval for mqtt 5

Kinplemelon:


src/i18n/Base.js
```diff
@@ -487,6 +487,10 @@ export default {
     zh: '未找到页面',
     en: 'This page could not be found',
   },
+  pageNotFound: {
+    zh: '页面不存在',
+    en: 'Not Fount',
```
ysfscream, line 492: Not Found?

src/views/Diagnose/components/WebSocketItem.vue
```diff
@@ -48,6 +48,34 @@
...
+            >
+              <BooleanSelect v-model="connection.clean" />
+            </el-form-item>
+          </el-col>
+          <el-col :span="8" v-if="connection.protocolversion === MQTT_V5_VALUE">
+            <el-form-item prop="sessionExpiryInterval" :label="$t('Tools.sessionExpiryInterval')">
+              <el-input
+                class="input-with-unit"
+                v-model.number="connection.sessionExpiryInterval"
+                :placeholder="$t('Tools.neverExpire')"
```
ysfscream, line 66: Please refer to https://github.com/emqx/MQTTX/blob/main/src/utils/mqttUtils.ts#L13

The never expire value is not default.

ysfscream (approved):

PR #669, opened 2022-11-24
feat(gateway): pop-up confirmation reminder before disabling the gateway

Kinplemelon:


src/i18n/Gateway.js
```diff
@@ -308,4 +308,8 @@ export default {
     zh: '设置',
     en: 'Settings',
   },
+  disableGatewayTip: {
+    zh: '禁用网关将导致通过此网关接入的连接全部断开，继续？',
+    en: 'Disabling the gateway will cause all connections accessed through this gateway to be disconnected. continue?',
```
ysfscream, line 313:
```suggestion
    en: 'Disabling the gateway will cause all connections accessed through this gateway to be disconnected. Continue?',
```

ysfscream (approved):

PR #679, opened 2022-11-28
refactor(global): re-write some logic to resolve warnings when compiling (WIP)

Kinplemelon:


Kinplemelon:
should I trans these components or files which use js to ts?

ysfscream:
> should I trans these components or files which use js to ts?

YES YOU CAN!

ysfscream (approved):

PR #687, opened 2022-12-02
refactor(limiter): refactor the limiter

Kinplemelon:


Kinplemelon:
There are some style problems here, which I will fix along with other style problems

ysfscream (approved):

PR #690, opened 2022-12-02
fix(auth): count the default value when create auth

Kinplemelon:


ysfscream:
Is it could be a standard function for Authn and Authz?

Kinplemelon:
> Is it could be a standard function for Authn and Authz?

less likely to be extracted into standard function

ysfscream:
> less likely to be extracted into standard function

Ok, fine. I thought they would be handled similarly and use different logic handlers.

ysfscream (approved):

PR #700, opened 2022-12-07
fix(http): reset response set after jump to login page

Kinplemelon:


Kinplemelon:
I am not sure this PR solved the problem which you said, please have a check after merge.

Kinplemelon:
Yes, I can't reproduce the problem you said in full, but I reproduced part of it, and it only didn't feedback if you clicked on login when you entered the wrong password, and then I fixed the problem.

ysfscream:
OK, this is a part of the issue which I have seen.

ysfscream (approved):

PR #751, opened 2022-12-19
refactor(rule): refactor create rule page

ysfscream:


src/components/InfoTooltip.vue
```diff
@@ -36,6 +36,7 @@ const tooltipPopperClass = computed(() => `info-tooltip ${props.popperClass}`)
   display: inline-block;
   font-size: 14px;
   cursor: pointer;
-  margin-left: 8px;
+  position: relative;
```
Kinplemelon, line 39: why change back to position relative😶‍🌫️

PR #785, opened 2023-01-03
refactor(bridge): remove disabled prop from some components that are not useful

Kinplemelon:


Kinplemelon:
I feel that my brain is not very clear, so I dealt with the problem of unifying the configuration page style first.

image

ysfscream (approved):

PR #786, opened 2023-01-04
refactor(mqtt): on-demand ingress/egress configs

ysfscream:
image

image

### DIFF
--- a/src/types/rule.d.ts
+++ b/src/types/rule.d.ts
@@ -157,7 +157,7 @@ export type MQTTIn = MQTTOut & {
 
 export type OtherBridge = Record<string, any>
 
-export type BridgeItem = HTTPBridge | MQTTOut | MQTTIn | OtherBridge
+export type BridgeItem = HTTPBridge | MQTTBridge | OtherBridge
 
 export interface ConnectorBase {
   name: string
--- a/src/components/TLSConfig/TLSBaseConfig.vue
+++ b/src/components/TLSConfig/TLSBaseConfig.vue
@@ -41,7 +41,6 @@ const record: WritableComputedRef<SSL> = computed({
 
 <style lang="scss">
 .TLS-base-config {
-  margin-bottom: 20px;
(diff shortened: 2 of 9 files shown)


src/types/rule.d.ts
```diff
@@ -217,3 +213,5 @@ export interface RuleMetrics {
 }
 
 export type RuleDataItemWithMetrics = Merge<RuleItem, { metrics: Metrics }>
+
+export type BridgeItem = HTTPBridge & MQTTBridge
```
Kinplemelon, line 217: The `&` operator doesn't seem very accurate here, but it seems to reduce the number of errors reported?🤔

src/types/rule.d.ts
```diff
@@ -155,10 +155,6 @@ export type MQTTIn = MQTTOut & {
   local_qos: QoSLevel
 }
 
-export type OtherBridge = Record<string, any>
-
-export type BridgeItem = HTTPBridge | MQTTOut | MQTTIn | OtherBridge
```
Kinplemelon, line 160: Why has otherBridge been removed here, is it meant to say that this type only exists on the enterprise branch?

src/types/rule.d.ts
ysfscream: I don't feel like I can say it's inaccurate, it's more likely to be a dynamic type here, but if it's a union type, it has to be asserted every time, but it's not fixed every time here either, it's based on the value of the API bounce back, so I didn't think of a way to solve the type for now and used `&`

src/types/rule.d.ts
ysfscream: I change back.

src/types/rule.d.ts
ysfscream: I change back.

ysfscream:
Something confused me

image

Kinplemelon:
> Something confused me
> 
> image

I will check it

PR #804, opened 2023-01-09
feat(overview): show license info and update license

Kinplemelon:
image
image
image


ysfscream:
Some styles should be optimized and can be going to the next PR. I will merge first.

ysfscream (approved):

PR #923, opened 2023-02-02
refactor(overview): replace async polling with sync polling

Kinplemelon:
please do not sync to the enterprise branch until release e5.0


Kinplemelon:
🤔 now I just changed the quest to get nodes data, would I use this logic for every request which needs polling?

ysfscream:
> 🤔 now I just changed the quest to get nodes data, would I use this logic for every request which needs polling?

I think so, especially for Overview page where there are a lot of polling operations, but for global requests like `alarms`, I don't think you can do that

PR #936, opened 2023-02-07
refactor(bridge influxdb): show tip when fill field value

Kinplemelon:
image


ysfscream:
@Kinplemelon How about moving the tooltip to the Fields label? To make it easier to explain?

image

Kinplemelon:
> @Kinplemelon How about moving the tooltip to the Fields label? To make it easier to explain?
> 
> image

The explanation here is specifically for the `value` field, would it be more appropriate to put it in the header of the `value` field?

ysfscream (approved):

PR #944, opened 2023-02-08
feat(mqtt extension): add tip for proxy sub

Kinplemelon:


src/i18n/Extension.js
```diff
@@ -309,4 +309,8 @@ export default {
     en: 'Enable to publish event message that client unsubscribed a topic successfully.',
     zh: '是否开启客户端已成功取消订阅主题事件消息。',
   },
+  proxySubTip: {
+    zh: '通过代理订阅为客户端订阅主题时，不会经过 AuthZ 检查权限。',
+    en: 'The authorization will not check authority when subscribing topics for clients through proxy subscription.',
```
ysfscream, line 314: Authorization won't be checked when clients subscribe to topics through Auto Subscribe.

ysfscream: https://www.emqx.io/docs/en/v5.0/mqtt/mqtt-auto-subscription.html#configuration

ysfscream: https://www.emqx.io/docs/en/v5.0/mqtt/mqtt-auto-subscription.html#configuration

ysfscream: https://www.emqx.io/docs/en/v5.0/mqtt/mqtt-auto-subscription.html#configuration

Kinplemelon: Authorization is the one which checks permission, it doesn't need to be checked?

The authorization will not check permission when clients subscribe to topics through Auto Subscribe.
this?

ysfscream: > The authorization will not check permission when clients subscribe to topics through Auto Subscribe.

OK

PR #980, opened 2023-02-15
fix(clients): show never expire

ysfscream:


src/common/constants.ts
```diff
@@ -213,3 +213,5 @@ export const AUTO_RESTART_INTERVAL_DEFAULT = '60s'
 export const ENCRYPTED_PWD_REG = /^\*{6}$/
 
 export const PASSWORD_REG = /^(?![\d]+$)(?![a-zA-Z]+$)(?![^\da-zA-Z]+$)[ -~]{8,64}$/
+
+export const SESSION_NEVER_EXPIRE_TIME = 4294967295
```
Kinplemelon, line 217: Would it be more readable to write `parseInt('0xFFFFFFFF', 16)` ?

ysfscream: YES. So nice. Thanks

PR #981, opened 2023-02-15
refactor(websocket): some change for notify

Kinplemelon:


Kinplemelon:
image
It seems a little ugly, maybe it will be optimized later

ysfscream (approved):

PR #1014, opened 2023-03-01
chore: remove root `request_timeout` from GCP PubSub bridge (e5.0)

thalesmg:
`resource_opts` already have such option, and keeping both could lead to inconsistent behavior.

Related PR: https://github.com/emqx/emqx/pull/10051

### DIFF
--- a/src/views/RuleEngine/Bridge/Components/BridgeConfig/BridgeHttpConfig.vue
+++ b/src/views/RuleEngine/Bridge/Components/BridgeConfig/BridgeHttpConfig.vue
@@ -172,7 +172,7 @@ export default defineComponent({
         pool_size: 4,
         enable_pipelining: 100,
         connect_timeout: '5s',
-        request_timeout: '5s',
+        request_timeout: '15s',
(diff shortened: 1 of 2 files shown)


src/views/RuleEngine/Bridge/Components/BridgeConfig/BridgeHttpConfig.vue
```diff
@@ -172,7 +172,7 @@ export default defineComponent({
         pool_size: 4,
         enable_pipelining: 100,
         connect_timeout: '5s',
-        request_timeout: '5s',
+        request_timeout: '15s',
```
thalesmg, line 175: We want to ensure `request_timeout` at the bridge root is the same as `request_timeout` inside `resource_opts`.  For now, making the default values match should help with quickly starting a new webhook bridge without having to fix this field.

thalesmg: Ideally, we probably should show only one of these fields (the root one) and copy that value to `resource_opts.request_timeout`.

PR #1034, opened 2023-03-07
refactor(http): add custom message func and use it when error occurre…

Kinplemelon:
…d with request

### DIFF
--- a/src/common/http.js
+++ b/src/common/http.js
@@ -1,5 +1,6 @@
 import axios from 'axios'
-import { ElMessage as M, ElNotification } from 'element-plus'
+import { ElNotification } from 'element-plus'
+import CustomMessage from '@/common/CustomMessage'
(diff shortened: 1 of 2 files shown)


src/common/http.js
```diff
@@ -107,7 +108,7 @@ axios.interceptors.response.use(
       const doNotPopupError = error.code === REQUEST_TIMEOUT_CODE && error.config.handleTimeoutSelf
       if (!respSet.has(0)) {
         if (!doNotPopupError) {
-          M.error('Some error occurred')
+          CustomMessage.error('Some error occurred')
```
ysfscream, line 111: `Some error occurred` should be an i18n translator.

ysfscream:
Could you please rename the `CustomMessage` func? It's not a sample way to call.

ysfscream (approved):

PR #1051, opened 2023-03-13
refactor: change kafka producer config structure

thalesmg:
https://emqx.atlassian.net/browse/EMQX-8386

In order to accomodate the changes from introducing Kafka Consumer as a separate bridge (https://github.com/emqx/emqx/pull/9564), the schema for the Kafka Producer changed a bit.

Originally, there was an intention for the Kafka Producer and Consumer bridges to be one entity like the MQTT Bridge is.  However, different from the MQTT bridge, Kafka drivers for consuming and producing are quite different, and configuration also becomes confusing by mixing those two types into one.  So, after internal discussions, we decided to separate those two.

We still have plans of renaming the Kafka Producer bridge type to `kafka_producer` to make it more semantic, but this'll require further changes in the backend and Hocon.


Kinplemelon:
Thank you for your contribution. The changes in this push request are not very complete. We will merge this PR and finish the missing changes after the backend PR has been merged.

PR #1088, opened 2023-03-22
refactor(enum): modify the logic of Kafka bridge type

Kinplemelon:


src/hooks/Rule/bridge/useBridgeTypeValue.ts
```diff
@@ -106,3 +106,30 @@ export const useBridgeTypeIcon = (): {
...
+      if (ingress) {
+        if (egress) {
+          return BridgeDirection.Both
+        }
+        return BridgeDirection.Ingress
+      }
+      return BridgeDirection.Egress
+    }
+
+    return BridgeDirection.Egress
```
ysfscream, line 129:
```js
const { ingress, egress } = bridge as MQTTBridge
    if (ingress && egress) {
      return BridgeDirection.Both
    } else if (ingress) {
      return BridgeDirection.Ingress
    }
  }
  return BridgeDirection.Egress
```

More better?

ysfscream (approved):

PR #1103, opened 2023-03-23
fix(kafka): modify the kafka consumer page

ysfscream:


src/components/ObjectArrayEditor.vue
```diff
@@ -15,12 +19,12 @@
     </el-table-column>
     <el-table-column width="100px">
       <template #header>
-        <el-button @click="addItem" size="small">
+        <el-button type="text" @click="addItem">
```
Kinplemelon, line 22: type `text` has been deprecated.
type a" has been deprecated.

Kinplemelon (approved):

Kinplemelon (approved):

src/components/ObjectArrayEditor.vue
ysfscream: Oh, yep. I found this in element UI doc

PR #1124, opened 2023-03-24
refactor(schema): use markdown content to show desc

Kinplemelon:
from 

image

to 

image

### DIFF
--- a/src/components/MarkdownContent.vue
+++ b/src/components/MarkdownContent.vue
@@ -12,17 +12,9 @@
   </el-row>
 </template>
 
-<script lang="ts">
-/* eslint-disable @typescript-eslint/ban-ts-comment */
-import { defineComponent, onUnmounted, PropType, watch } from 'vue'
-
-export default defineComponent({
-  name: 'MarkdownContent',
-})
-</script>
-
 <script setup lang="ts">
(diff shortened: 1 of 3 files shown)


src/components/MarkdownContent.vue
```diff
@@ -12,17 +12,9 @@
   </el-row>
 </template>
 
-<script lang="ts">
```
Kinplemelon, line 15: I don't know why, but it caused an incomprehensible error in the schema form here. I tried for a while but couldn't solve it, so I just deleted it...

Kinplemelon: Type '{ content: string; }' is not assignable to type 'IntrinsicAttributes & Partial<{}> & Pick<Readonly<{} & {} & {}> & VNodeProps & AllowedComponentProps & ComponentCustomProps, "style" | ... 10 more ... | "class">'.
  Property 'content' does not exist on type 'IntrinsicAttributes & Partial<{}> & Pick<Readonly<{} & {} & {}> & VNodeProps & AllowedComponentProps & ComponentCustomProps, "style" | ... 10 more ... | "class">'.

ysfscream: Where is the code?

ysfscream (approved):

PR #1371, opened 2023-05-17
fix(config): update error tip

ysfscream:
https://emqx.atlassian.net/browse/ED-619


Kinplemelon:
Forgot to check after machine translation😵

PR #1448, opened 2023-06-06
refactor(bridge): update `request_timeout`

Kinplemelon:


thalesmg:
Just a note: a related change to this is also that `auto_restart_interval` should be dropped from all resources. (ref.: https://github.com/emqx/emqx/pull/10910)

ysfscream (approved):

PR #1542, opened 2023-06-26
Fix bugs

ysfscream:
https://emqx.atlassian.net/browse/ED-770
https://emqx.atlassian.net/browse/ED-768
https://emqx.atlassian.net/browse/ED-769
https://emqx.atlassian.net/browse/ED-753

### DIFF
--- a/src/components/Oneof.vue
+++ b/src/components/Oneof.vue
@@ -166,10 +166,9 @@ const currentIns = getCurrentInstance()
  * form validation after setting modelValue to null following enable.
  */
 watch(isDisabled, async (val) => {
-  if (!val) {
+  if (val) {
     const formItem = props.formItem || currentIns?.parent?.exposed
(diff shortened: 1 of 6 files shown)


src/components/Oneof.vue
```diff
@@ -166,10 +166,9 @@ const currentIns = getCurrentInstance()
  * form validation after setting modelValue to null following enable.
  */
 watch(isDisabled, async (val) => {
-  if (!val) {
+  if (val) {
     const formItem = props.formItem || currentIns?.parent?.exposed
     if (isFunction(formItem?.clearValidate)) {
-      await nextTick()
```
Kinplemelon, line 172: Removing here may cause a little problem. Deleting here is to solve the https://emqx.atlassian.net/browse/ED-753, right? Let's take a look when I finish processing the webhook😵

PR #1568, opened 2023-06-29
fix(mongo): remove `topology.pool_size`

thalesmg:
See https://github.com/emqx/emqx/pull/11163

image

### DIFF
--- a/src/hooks/Rule/bridge/useSchemaBridgePropsLayout.ts
+++ b/src/hooks/Rule/bridge/useSchemaBridgePropsLayout.ts
@@ -101,7 +101,6 @@ export default (
           'collection',
           'ssl',
           'payload_template',
-          'topology.pool_size',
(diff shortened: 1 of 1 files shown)


Kinplemelon:
Why is the style on your computer like this? 😱 May I inquire as to which browser you are using?

thalesmg:
> Why is the style on your computer like this? 😱 May I inquire as to which browser you are using?

:joy_cat: 
I'm using Firefox 114.0.2

thalesmg:
Aha, this might be throwing you off (it might have been on when I took the screenshot): https://darkreader.org/help/en/ :dark_sunglasses:

Kinplemelon:
Thank you for your reply! I tried to reproduce it using Firefox + Dark Reader, and it seems to be normal, so I have decided to leave it as it is for now.🤣
image

PR #1650, opened 2023-07-26
feat(login): support cancel request when logout

ysfscream:
https://emqx.atlassian.net/browse/ED-818

### DIFF
--- a/src/store/index.ts
+++ b/src/store/index.ts
@@ -50,6 +50,7 @@ export default createStore({
     schemaStoreMap: new Map(),
     ruleEventList: [] as Array<RuleEvent>,
     ruleEventRequest: undefined as undefined | Promise<any>,
+    abortControllers: [] as AbortController[],
   },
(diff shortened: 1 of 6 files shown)


src/store/index.ts
```diff
@@ -129,6 +130,22 @@ export default createStore({
     SET_RULE_EVENT_REQUEST(state, payload: Promise<any>) {
       state.ruleEventRequest = payload
     },
+    ADD_ABORT_CONTROLLER(state, controller) {
+      state.abortControllers.push(controller)
+    },
+    CLEAR_ABORT_CONTROLLERS(state) {
+      console.log(state.abortControllers)
```
Kinplemelon, line 137: 👀

Kinplemelon (approved):

src/store/index.ts
ysfscream: Removed. 0.0

PR #1782, opened 2023-08-24
fix(webhook): judge the type of trigger

Kinplemelon:


src/views/Webhook/components/Trigger.vue
```diff
@@ -295,17 +306,34 @@ const validate = () => {
...
 const selectedType = computed({
   get() {
     if (!fromDataArr.value.length && isSelectedEvent.value) {
       return TriggerType.Events
     }
-    let msgEventNum = trueTopicList.value.length
-      ? selectedMsgEventList.value.length - 1 + trueTopicList.value.length
-      : selectedMsgEventList.value.length
-    const isAllMsgEvent = msgEventNum === fromDataArr.value.length
-    if (isAllMsgEvent) {
```
Kinplemelon, line 308: Is a problem with the logic of judging whether the trigger type is message. I can't understand why I wrote it like this before...

ysfscream: This is growth! 🚀

ysfscream (approved):

PR #1824, opened 2023-08-31
refactor(webhook): change prefix to suffix

Kinplemelon:


ysfscream:
Why change these?

Kinplemelon:
> Why change these?

because our PM said the prefix is too long and influence user to look in list or flow page

ysfscream (approved):

PR #1834, opened 2023-09-05
fix(webhook): fix webhook name in stats page

ysfscream:
https://emqx.atlassian.net/browse/ED-942

### DIFF
--- a/src/views/RuleEngine/Bridge/BridgeDetail.vue
+++ b/src/views/RuleEngine/Bridge/BridgeDetail.vue
@@ -3,7 +3,10 @@
     <div class="detail-top">
       <detail-header
         v-if="!isFromRule"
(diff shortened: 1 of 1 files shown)


src/views/RuleEngine/Bridge/BridgeDetail.vue
```diff
@@ -184,6 +187,14 @@ const isWebhook = computed(() => {
   return route.name === 'webhook-detail-stats'
 })
 
+const bridgeName = computed(() => {
+  const { name } = bridgeInfo.value
+  if (name && isWebhook.value) {
+    return name.replace(WEBHOOK_SUFFIX, '')
```
Kinplemelon, line 193: can we use https://github.com/emqx/emqx-dashboard5/blob/01ea27d5f4daba445592276d2842a3b434a17515/src/hooks/Webhook/useWebhookUtils.ts#L52

PR #1929, opened 2023-09-28
refactor(rule): reuse re pub form in outputs drawer

Kinplemelon:
https://emqx.atlassian.net/browse/ED-972


src/i18n/RuleEngine.js
```diff
@@ -638,6 +638,34 @@ export default {
...
+    en: 'Payload Format Indicator',
+  },
+  messageExpiryInterval: {
+    zh: '遗嘱消息过期时间',
+    en: 'Message Expiry Interval',
+  },
+  contentType: {
+    zh: '遗嘱消息描述',
+    en: 'Content Type',
+  },
```
ysfscream, line 656: It seems this has nothing to do with the last will message; I believe you might have referred to the wrong translation.

Kinplemelon: oh, let me see

ysfscream (approved):

PR #2006, opened 2023-11-16
refactor(gateway): convert the old syntax to vue setup script

ysfscream:
https://emqx.atlassian.net/browse/ED-1049


src/views/General/SSODetail.vue
```diff
@@ -37,7 +37,8 @@ import {
   EmqxDashboardSsoLdapLdap,
 } from '@/types/schemas/dashboardSingleSignOn.schemas'
 import { ElMessage } from 'element-plus'
-import { Component, ComputedRef, Ref, computed, ref } from 'vue'
+import type { Component, ComputedRef, Ref } from 'vue'
```
Kinplemelon, line 40: 😲

ysfscream: Yes, I saw an 'Export Component' error in the console, so I tried to fix it with 'import type,' and it worked...

Kinplemelon (approved):

PR #2012, opened 2023-11-17
feat(gateway): adapt the jt808 gateway

ysfscream:
https://emqx.atlassian.net/browse/ED-1048

### DIFF
--- a/src/i18n/Gateway.js
+++ b/src/i18n/Gateway.js
@@ -366,6 +366,6 @@ export default {
   },
   dnTopic: {
     zh: '下行主题',
-    en: 'Dn Topic',
+    en: 'Down Topic',
   },
 }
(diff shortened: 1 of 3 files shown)


src/i18n/Gateway.js
```diff
@@ -344,4 +344,28 @@ export default {
...
+    zh: '认证',
+    en: 'Authentication',
+  },
+  upTopic: {
+    zh: '上行主题',
+    en: 'Up Topic',
+  },
+  dnTopic: {
+    zh: '下行主题',
+    en: 'Dn Topic',
```
Kinplemelon, line 369: Can `down` be abbreviated to `dn`?

ysfscream: yep

ysfscream: fixed

PR #2052, opened 2023-11-24
refactor(metrics): support for dynamically multi-level charts render

ysfscream:


src/components/Metrics/OverviewMetrics.vue
```diff
@@ -61,7 +65,7 @@
             </div>
             <div class="metric-bar-bd">
               <!-- Bar Chart -->
-              <div class="bar-container" :id="createRandomString()" ref="BarChartEle"></div>
+              <div class="bar-container" id="bar-chart-rate"></div>
```
Kinplemelon, line 68: remember the battle when I just came here...

PR #2116, opened 2023-12-05
fix(gateway): update the layout and values

ysfscream:
https://emqx.atlassian.net/browse/ED-1072


src/views/Gateway/components/coapBasic.vue
```diff
@@ -64,6 +56,11 @@
             <el-input v-model="cValue.mountpoint" :placeholder="cValueDefault.mountpoint" />
           </el-form-item>
         </el-col>
+        <el-col :span="24">
```
Kinplemelon, line 59: should these sync to the dev branch?

PR #2140, opened 2023-12-07
fix(api key): disabled role select when viewing

Kinplemelon:
don't disabled the whole form just because the copy button need to be enabled....


ysfscream (approved):

scripts/checkPermission.js
ysfscream, line 1: Excellent script; Can we add it to the CI test process?

PR #2182, opened 2023-12-13
fix(action): fixed the issue that can not create redis

Kinplemelon:
https://emqx.atlassian.net/browse/ED-1114

### DIFF
--- a/src/hooks/Rule/useDataHandler.ts
+++ b/src/hooks/Rule/useDataHandler.ts
@@ -209,9 +209,12 @@ export const useBridgeDataHandler = (): {
   const { splitBySpace, transCommandArrToStr } = useRedisCommandCheck()
   const handleRedisBridgeData = async (bridgeData: any) => {
     try {
-      bridgeData.command_template = await splitBySpace(bridgeData.command_template)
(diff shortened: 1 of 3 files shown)


src/hooks/Rule/useDataHandler.ts
```diff
@@ -209,9 +209,12 @@ export const useBridgeDataHandler = (): {
   const { splitBySpace, transCommandArrToStr } = useRedisCommandCheck()
   const handleRedisBridgeData = async (bridgeData: any) => {
     try {
-      bridgeData.command_template = await splitBySpace(bridgeData.command_template)
+      if (bridgeData.command_template && typeof bridgeData.command_template === 'string') {
+        bridgeData.command_template = await splitBySpace(bridgeData.command_template)
+      }
       return bridgeData
     } catch (error) {
+      console.error(error)
```
ysfscream, line 217: Remove console?

ysfscream (approved):

src/hooks/Rule/useDataHandler.ts
Kinplemelon: I feel it’s better to print it out if something goes error, otherwise it will be difficult to find out where produce error 🤔️

Issue #2202, opened 2023-12-15
use default name for favicon.ico


steffstefferson:
In our log we see that some browsers (especially chrome) try to access /favicon.ico which does not exists. The ressource is not found and the following error is written to our log:
`2023-12-13T06:32:31.496599+00:00 [warning] msg: unexpected_api_access, mfa: emqx_dashboard_not_found:init/2, line: 25, request: #{bindings => #{},body_length => 0,cert => undefined,has_body => false,headers => #{<<"accept-encoding">> => <<"gzip, deflate, br">>,<<"accept-language">> => <<"en-GB,en-US;q=0.9,en;q=0.8">>,<<"host">> => <<"mqtt-broker-ui-prod.mqtt.aws.pnetcloud.ch">>,<<"sec-fetch-dest">> => <<"empty">>,<<"sec-fetch-mode">> => <<"no-cors">>,<<"sec-fetch-site">> => <<"none">>,<<"user-agent">> => <<"Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/119.0.0.0 Safari/537.36">>,<<"x-amzn-trace-id">> => <<"Root=1-65794fff-695b85f4353bbeee622487b7">>,<<"x-forwarded-for">> => <<"172.x.x.x">>,<<"x-forwarded-port">> => <<"443">>,<<"x-forwarded-proto">> => <<"https">>},host => <<"mqtt-broker-ui-prod.mqtt.aws.pnetcloud.ch">>,host_info => undefined,method => <<"GET">>,path => <<"/favicon.ico">>,path_info => undefined,peer => {{10,x,x,x},22088},pid => <0.13431.157>,port => 80,qs => <<>>,ref => 'http:dashboard',scheme => <<"http">>,sock => {{10,xx,x},18083},streamid => 1,version => 'HTTP/1.1'}`

For me it is unclear why they do this because actually the favicon is correctly defined in the index.html. Maybe when adding a bookmark? 
It is not a curcial error, but since we strive for a zero-errors in production we stumble over this error from time to time.

So the easiest mitigation is to rename emq.ico to favicon.ico, see my pr.

PR #2203, opened 2023-12-15
fix(index): default favicon name

steffstefferson:
https://github.com/emqx/emqx-dashboard5/issues/2202


Kinplemelon:
Thanks for your contribution! We will merge this PR into branch dev/1.6.0, and the fix will be included in the next version.

PR #2272, opened 2024-01-04
refactor(schema): add rule for type integer and add limit in number input

Kinplemelon:


src/views/RuleEngine/components/RuleForm.vue
```diff
@@ -184,7 +184,7 @@ enum RightTab {
   Sources,
   Actions,
 }
-const rightBlockActiveTab = ref(RightTab.Sources)
```
Kinplemelon, line 187: This was previously set to `source` for easy viewing during development...

ysfscream (approved):

PR #2376, opened 2024-01-23
fix(action): remove qos2 for ingress in mqtt source

Kinplemelon:
https://emqx.atlassian.net/browse/ED-1224


Kinplemelon:
cloud don't use the source, so...

ysfscream (approved):

PR #2379, opened 2024-01-24
fix(doc): add batch settings documentation links

ysfscream:
https://emqx.atlassian.net/browse/ED-1190

### DIFF
--- a/src/components/BatchSettings.vue
+++ b/src/components/BatchSettings.vue
@@ -18,7 +18,11 @@
             <div>{{ $t('Base.downloadTemplatePrompt') }}</div>
             <ul>
               <li>{{ $t('Base.batchSettingDownloadFirst') }}</li>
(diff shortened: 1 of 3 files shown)


src/components/BatchSettings.vue
```diff
@@ -18,7 +18,7 @@
             <div>{{ $t('Base.downloadTemplatePrompt') }}</div>
             <ul>
               <li>{{ $t('Base.batchSettingDownloadFirst') }}</li>
-              <li>{{ $t('Base.moreImportInstructions') }}</li>
+              <li v-html="$t('Base.moreImportInstructions', { link: checkDocMap[type] })"></li>
```
Kinplemelon, line 21: can we use https://vue-i18n.intlify.dev/guide/advanced/component.html#basic-usage ?

ysfscream: Sure, this is because I am too lazy to use it. 😅

ysfscream: Updated

PR #2465, opened 2024-03-01
feat(schema registry): support type JSON

Kinplemelon:


ysfscream:
image

Perhaps we should consider adding support for the `json` language in this editor when the user selects the JSON type.

Kinplemelon:
> image
> Perhaps we should consider adding support for the `json` language in this editor when the user selects the JSON type.

oh yes!

ysfscream (approved):

PR #2474, opened 2024-03-05
refactor(connector & source): remove some useless code

Kinplemelon:


Kinplemelon:
I need to merge it into the enterprise version and then look at the code.

PR #2479, opened 2024-03-06
fix(rule): error message when test fails

Kinplemelon:
https://emqx.atlassian.net/browse/ED-1268

### DIFF
--- a/src/api/ruleengine.ts
+++ b/src/api/ruleengine.ts
@@ -125,7 +125,11 @@ export function deleteRules(id: string): Promise<any> {
 
 export function testsql(body: Record<string, unknown>): Promise<any> {
   return http.post('/rule_test', body, {
-    transformResponse: [(data: string) => data],
+    transformResponse: [
+      (data: string, responseHeader: any, code: number) => {
+        return code > 299 ? JSON.parse(data) : data
+      },
+    ],
   })
 }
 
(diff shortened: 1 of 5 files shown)


src/api/sources.ts
```diff
@@ -6,7 +7,14 @@ export const getSources = async (): Promise<Array<Source>> => {
   try {
     const data = await http.get(`/sources`)
     return Promise.resolve(
-      data.map((item: Omit<Source, 'id'>) => ({ id: getBridgeKey(item as any), ...item })),
+      data.map((item: Omit<Source, 'id'>) => {
+        const id = getBridgeKey(item as any)
```
ysfscream, line 11: Should we rename this function name? Like getConnectorKey?

ysfscream (approved):

src/api/ruleengine.ts
```diff
@@ -125,7 +125,11 @@ export function deleteRules(id: string): Promise<any> {
 
 export function testsql(body: Record<string, unknown>): Promise<any> {
   return http.post('/rule_test', body, {
-    transformResponse: [(data: string) => data],
+    transformResponse: [
+      (data: string, responseHeader: any, code: number) => {
+        return code > 299 ? JSON.parse(data) : data
```
ysfscream, line 130: Give the 299 a clear name, or leave some comments here to inform others about the number.

PR #2506, opened 2024-03-14
test(ci): add ci for check audit dict

Kinplemelon:
if fails

image


ysfscream (commented):
It seems like you're already a professional CI expert. 👀

PR #2523, opened 2024-03-20
feat(global): support get token from query

Kinplemelon:
`c` is customer, I don't want it to be too obvious


ysfscream (approved):

ysfscream:
@Kinplemelon Good idea! 💡

PR #2525, opened 2024-03-21
feat(layout): support hide menu from query

Kinplemelon:
correct the data handler

### DIFF
--- a/src/i18n/General.js
+++ b/src/i18n/General.js
@@ -260,7 +260,7 @@ export default {
     en: 'Disabled',
   },
   SSOEnable: {
-    zh: '启用 {backend} SSO',
+    zh: '启用{backend} SSO',
     en: 'Enable {backend} SSO',
   },
   baseDN: {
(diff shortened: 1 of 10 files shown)


src/views/General/components/SSOForm/SAMLForm.vue
```diff
@@ -8,7 +8,7 @@
     :label-width="180"
     :validate-on-rule-change="false"
   >
-    <el-form-item prop="enable" :label="tl('SSOEnable', { backend: 'SAML' })">
+    <el-form-item prop="enable" :label="tl('SSOEnable', { backend: ' SAML' })">
```
ysfscream, line 11: Why add a blank to the `SAML`?

src/views/General/components/SSOForm/LDAPForm.vue
```diff
@@ -7,7 +7,7 @@
     :rules="rules"
     :label-width="150"
   >
-    <el-form-item prop="enable" :label="tl('SSOEnable', { backend: 'LDAP' })">
+    <el-form-item prop="enable" :label="tl('SSOEnable', { backend: ' LDAP' })">
```
ysfscream, line 10: same here.

ysfscream (approved):

src/i18n/General.js
```diff
@@ -260,7 +260,7 @@ export default {
     en: 'Disabled',
   },
   SSOEnable: {
-    zh: '启用 {backend} SSO',
+    zh: '启用{backend} SSO',
```
Kinplemelon, line 263: Because a space was removed here; the reason for removing the space is that when the `backend` is in Chinese, it's not appropriate to have a space here.

PR #2534, opened 2024-03-26
fix(sso): chinese of sso

Kinplemelon:
https://emqx.atlassian.net/browse/ED-1321

https://emqx.atlassian.net/browse/ED-1322


Kinplemelon:
I will merge it first

PR #2587, opened 2024-04-18
refactor(rule): support `is_template`

Kinplemelon:


src/components/SchemaFormItem.tsx
```diff
@@ -135,9 +142,20 @@ export default defineComponent({
                   v-model={formItemValue.value}
                   lang="sql"
                   disabled={isDisabled}
+                  {...customProps}
                 />
               </div>
             )
+          } else if (isTemplate) {
```
ysfscream, line 149: Since the value is already returned in the above `if` statement, it is recommended to use `if` here instead of `else if`.

ysfscream (approved):

emqxqa:
[Test Execution]()

PR #2643, opened 2024-05-10
fix(authz): disabled top and bottom item move buttons

ysfscream:
https://emqx.atlassian.net/browse/ED-1364
https://emqx.atlassian.net/browse/ED-1375
https://emqx.atlassian.net/browse/ED-1376

### DIFF
--- a/src/common/http.js
+++ b/src/common/http.js
@@ -112,12 +112,16 @@ axios.interceptors.response.use(
           Array.isArray(error.config.errorsHandleCustom) &&
           error.config.errorsHandleCustom.includes(status)
         if (!handleErrorSelf) {
-          if (data?.code === NAME_PWD_ERROR) {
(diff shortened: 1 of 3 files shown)


src/common/http.js
```diff
@@ -112,12 +112,16 @@ axios.interceptors.response.use(
           Array.isArray(error.config.errorsHandleCustom) &&
           error.config.errorsHandleCustom.includes(status)
         if (!handleErrorSelf) {
```
ysfscream, line 114: This file is an original JavaScript file, so I temporarily removed the TypeScript syntax, util we move it to the ts file.

Kinplemelon: 🧐

Kinplemelon (approved):

emqxqa:
[Test Execution]()

PR #2658, opened 2024-05-13
fix(rule): handling the case where `trigger_times` has a length greater than 1

Kinplemelon:
https://emqx.atlassian.net/browse/ED-1380


ysfscream (approved):

emqxqa:
[TestExecution]()

PR #2661, opened 2024-05-14
fix(rule): hide source in `FromSelect`

Kinplemelon:
https://emqx.atlassian.net/browse/ED-1384

https://emqx.atlassian.net/browse/ED-1383


ysfscream (approved):

emqxqa:
[TestExecution]()

PR #2692, opened 2024-05-22
fix(rule): Remove logs that are not rule executions and are without action info

Kinplemelon:
https://emqx.atlassian.net/browse/ED-1402


ysfscream (approved):

emqxqa:
[TestExecution]()

PR #2703, opened 2024-05-28
refactor(action): deal with the schema for s3

Kinplemelon:


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-3482)

PR #2707, opened 2024-05-28
fix(schema form): handling of rules after selecting oneof ref

Kinplemelon:
for s3


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-2320)

PR #2748, opened 2024-06-13
fix(client): some texts cannot be displayed completely when they are too long

Kinplemelon:
https://emqx.atlassian.net/browse/ED-1359

https://emqx.atlassian.net/browse/ED-465


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-2471)

ysfscream (approved):

PR #2753, opened 2024-06-13
fix(authz): incomplete display when topic is too long

Kinplemelon:
https://emqx.atlassian.net/browse/ED-1212


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-2481)

PR #2758, opened 2024-06-18
fix(listener): delete cacertfile when it is empty on quic

ysfscream:
https://emqx.atlassian.net/browse/ED-1438


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-2587)

PR #2759, opened 2024-06-19
fix(plugin): do not request config api with old version pkg

ysfscream:
https://emqx.atlassian.net/browse/ED-1440


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-2613)

Kinplemelon (approved):

PR #2788, opened 2024-07-04
fix(client): handling data duplication when the list is too short

Kinplemelon:
https://emqx.atlassian.net/browse/ED-1458


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-2826)

ysfscream (approved):

PR #2791, opened 2024-07-05
fix(overview): change the default name for save chart image

ysfscream:


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-2832)

Kinplemelon (approved):

PR #2794, opened 2024-07-05
fix(client): display of message queue info when it is too long

Kinplemelon:
https://emqx.atlassian.net/browse/ED-1460


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-2872)

ysfscream (approved):

PR #2795, opened 2024-07-05
fix(sso): correct chinese label in oidc

Kinplemelon:
https://emqx.atlassian.net/browse/ED-1462


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-2873)

ysfscream (approved):

PR #2806, opened 2024-07-10
Sync code from enterprise

Kinplemelon:


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-2929)

PR #2814, opened 2024-07-12
fix(authn): set headers when type is JWKS

Kinplemelon:
https://emqx.atlassian.net/browse/ED-1476


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-2963)

PR #2828, opened 2024-07-19
fix(monitoring): remove datadog

ysfscream:


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-3069)

PR #2830, opened 2024-07-22
refactor(hot upgrade): set special timeout for some request

Kinplemelon:
Could I just unlimit the timeout for these requests?

### DIFF
--- a/src/api/hotUpgrade.ts
+++ b/src/api/hotUpgrade.ts
@@ -2,13 +2,13 @@ import http from '@/common/http'
 import { NodeUpgradeData, HotUpgradePackage } from '@/types/typeAlias'
 
 export const upgradeNode = (node: string): Promise<void> => {
-  return http.post(`/relup/upgrade/${encodeURIComponent(node)}`)
+  return http.post(`/relup/upgrade/${encodeURIComponent(node)}`, { timeout: 600000 })
 }
 
 export const uploadUpgradePackage = (file: Blob): Promise<void> => {
   const formData = new FormData()
   formData.append('plugin', file)
-  return http.post(`/relup/package/upload`, formData)
+  return http.post(`/relup/package/upload`, formData, { timeout: 600000 })
 }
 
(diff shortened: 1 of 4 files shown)


ysfscream:
@Kinplemelon YES, you can set the timeout to unlimited, but I recommend setting it to a large value. If users encounter a network error, they will wait for the request for a long time or indefinitely. It is not good for users to wait so long to discover an issue. Do you think this is correct?

Kinplemelon:
Wouldn't a network error cause the request to fail?

ysfscream:
Indeed, some network errors can cause requests to fail, but there are also special cases where requests might hang indefinitely without failing. e.g.

1. Connection Timeout:
Example: Trying to connect to a closed port without a timeout.
```python
socket.connect(('example.com', 12345))  # May hang indefinitely
```

2. Read Timeout:
Example: Requesting from an extremely slow API without a timeout.
```python
requests.get('https://very-slow-api.com')  # May not respond for a long time
```

3. Network Packet Loss:
Severe packet loss can cause continuous TCP retransmissions. This is usually a network environment issue, difficult to simulate in code.

4. DNS Resolution Issues:
Example: Attempting to resolve a non-existent domain with unresponsive DNS servers.
```python
socket.gethostbyname('nonexistent-domain.com')  # May get stuck
```

5. Proxy Server Problems:
Example: Using a non-existent proxy.
```python
requests.get('http://example.com', proxies={'http': 'http://non-existent-proxy:8080'})
```

6. Firewall Anomalies:
Typically a network configuration issue, not directly visible in application code.

These scenarios can all lead to requests hanging rather than failing immediately. Solutions include setting appropriate timeouts, implementing retry mechanisms, and using the circuit breaker pattern.

Kinplemelon:
Oh, thanks for the detailed reply, I'll change the timeout

ysfscream (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-3489)

PR #2832, opened 2024-07-24
fix(rule): check polling status before getting node logs

Kinplemelon:
https://emqx.atlassian.net/browse/ED-1485

Thanks @0721Betty for finding this problem and offering the solution!


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-3185)

ysfscream (approved):

PR #2837, opened 2024-07-26
fix(msg transform): corrects attrs that can be used

Kinplemelon:
https://emqx.atlassian.net/browse/ED-1487


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-3201)

ysfscream (approved):

PR #2851, opened 2024-08-01
fix(flow): function node form action column display error

Kinplemelon:
https://emqx.atlassian.net/browse/ED-1344


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-3269)

ysfscream (approved):

PR #2853, opened 2024-08-02
fix(rule): bug when deleting input

Kinplemelon:
https://emqx.atlassian.net/browse/ED-1506


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-3291)

ysfscream (approved):

PR #2860, opened 2024-08-06
Hide OIDC and Message Transform

Kinplemelon:


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-3349)

PR #2862, opened 2024-08-07
fix(action): remove `rules` field when submitting

Kinplemelon:
https://emqx.atlassian.net/browse/ED-1465


ysfscream (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-3576)

PR #2866, opened 2024-08-08
fix(authz): update ignore desc

ysfscream:
https://emqx.atlassian.net/browse/ED-1512


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-3577)

PR #2872, opened 2024-08-09
fix(topics): button that shouldn't be there

Kinplemelon:
https://emqx.atlassian.net/browse/ED-1515


ysfscream (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-3581)

PR #2878, opened 2024-08-12
fix(icon): update favicon

ysfscream:
https://emqx.atlassian.net/browse/ED-1505

image


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-3459)

PR #2898, opened 2024-08-20
fix(authz): error after failed submission

Kinplemelon:
https://emqx.atlassian.net/browse/ED-1547

### DIFF
--- a/src/views/Auth/components/AuthzManager.vue
+++ b/src/views/Auth/components/AuthzManager.vue
@@ -260,7 +260,7 @@ import {
   updateBuiltInDatabaseData,
   updateAllBuiltInDatabaseData,
 } from '@/api/auth'
-import _ from 'lodash'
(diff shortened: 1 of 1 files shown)


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-3602)

src/views/Auth/components/AuthzManager.vue
```diff
@@ -260,7 +260,7 @@ import {
   updateBuiltInDatabaseData,
   updateAllBuiltInDatabaseData,
 } from '@/api/auth'
-import _ from 'lodash'
+import _, { cloneDeep } from 'lodash'
```
ysfscream, line 263: Keep styles. If any place uses the `_.xxx` function, only replace to use the `xxx`.

ysfscream (approved):

PR #2901, opened 2024-08-20
fix(flow): blob storage read-only display problem

Kinplemelon:
https://emqx.atlassian.net/browse/ED-1549


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-3605)

ysfscream (approved):

PR #2902, opened 2024-08-20
fix(listener): add help-text for rate limiter

ysfscream:
https://emqx.atlassian.net/browse/ED-1540


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-3607)

PR #2904, opened 2024-08-20
fix(msg transform): prop not being cleared

Kinplemelon:
https://emqx.atlassian.net/browse/ED-1553

https://emqx.atlassian.net/browse/ED-1527


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-3609)

Kinplemelon:
Added desc to metric item `queueing`, `inflight` and `routes`, where the desc for item `queueing` and `inflight` is the same as they are in the action.

ysfscream (approved):

PR #2906, opened 2024-08-21
fix(auth): fix principal tips format error

ysfscream:
https://emqx.atlassian.net/browse/ED-1541


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-3740)

PR #2907, opened 2024-08-21
fix(hot upgrade): request that cannot be canceled

Kinplemelon:
https://emqx.atlassian.net/browse/ED-1555


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-3618)

PR #2913, opened 2024-08-22
fix(listener): update rate limiter description

ysfscream:
https://emqx.atlassian.net/browse/ED-1559


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-3745)

PR #2915, opened 2024-08-22
fix(plugin): fix info can not shown totally

ysfscream:
https://emqx.atlassian.net/browse/ED-1563


Kinplemelon (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-3748)

PR #2917, opened 2024-08-22
fix(auth): fix missing kerberos config on settings

ysfscream:
https://emqx.atlassian.net/browse/ED-1567


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-3640)

Kinplemelon (approved):

PR #2923, opened 2024-08-26
fix(action): issue when creating datalayers action

Kinplemelon:
https://emqx.atlassian.net/browse/ED-1574


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-3674)

ysfscream (approved):

PR #2925, opened 2024-08-26
fix(plugin): fix can not show the no-config plugin ui

ysfscream:
https://emqx.atlassian.net/browse/ED-1579


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-3721)

PR #2932, opened 2024-08-28
refactor(connector): adapt new configs in iotdb

Kinplemelon:
https://emqx.atlassian.net/browse/ED-1573

https://emqx.atlassian.net/browse/ED-1583

### DIFF
--- a/src/hooks/Rule/bridge/useBridgeTypeValue.ts
+++ b/src/hooks/Rule/bridge/useBridgeTypeValue.ts
@@ -300,6 +300,7 @@ export const useConnectorDirection = (): {
 
 export const useConnectorSchema = (): {
   getTypeRefKey: (type: string) => string
+  typeWithMultipleRefKeyMap: Map<BridgeType, Array<string>>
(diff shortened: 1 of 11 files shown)


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-3732)

Kinplemelon:
@0721Betty FYI. e5.8.1 will add a new schema that can be used to create IoTDB type connector
image

0721Betty:
> @0721Betty FYI. e5.8.1 will add a new schema that can be used to create IoTDB type connector image

Why does IoTDB need two root refs? Can't it be merged into one root ref like mongoDB/redis?

Kinplemelon:
> > @0721Betty FYI. e5.8.1 will add a new schema that can be used to create IoTDB type connector image
> 
> Why does IoTDB need two root refs? Can't it be merged into one root ref like mongoDB/redis? If they are very different, need to be split into two connectors like kafka?

The developers tried to merge into one schema, then found that there was no way to handle the old data after the merge, so they finally split it into two schemas.

ysfscream (approved):

PR #2942, opened 2024-09-03
fix: confirm tip for reset license

JimMoen:


.tool-versions
```diff
@@ -0,0 +1 @@
+nodejs 16.20.2
```
ysfscream, line 1: I'm not quite sure what the purpose of this Node.js version declaration is here?

JimMoen: Manage multiple runtime versions by
https://asdf-vm.com/

ysfscream: COOL, it looks good, but I've already set the .nvmrc to change the Node.js version automatically, and `v16.20.2` is a specific version number. Most of us might not necessarily have this exact version. For example, mine is `v16.14.2`

ysfscream: Or let's do this: ignore this modification for now and use it locally. Because I currently don't want to add duplicate configuration files to the project, I need to maintain consistency.

I'll look at this feature to see if we need to switch to this tool. If it works, we'll add it back for everyone to use. After the modification, I'll approve this PR. Thank you very much for your contribution.

JimMoen: `.tool-versions` reverted  and ignored.

ysfscream (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-4104)

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-4105)

PR #2954, opened 2024-09-14
fix(snowflake): update the private key component

ysfscream:
https://emqx.atlassian.net/browse/ED-1603


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-4279)

PR #2958, opened 2024-09-19
test(schemas): remove ignore item from prettier and eslint config

Kinplemelon:
I have found that ignoring schemas causes a large number of changes to the schemas...


ysfscream (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-6846)

PR #2960, opened 2024-09-19
fix(sql): fix can not handle big float number

ysfscream:
https://emqx.atlassian.net/browse/ED-1601

https://github.com/emqx/emqx/issues/13785


Kinplemelon:
So does this PR need to be updated to use `shared-utils`?

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-4285)

PR #2965, opened 2024-09-20
refactor(global): use dayjs instead of moment

Kinplemelon:
https://emqx.atlassian.net/browse/ED-1585

https://emqx.atlassian.net/browse/ED-1586


ysfscream (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-4091)

PR #2967, opened 2024-09-23
refactor(http): convert to ts

Kinplemelon:
https://emqx.atlassian.net/browse/ED-1585

https://emqx.atlassian.net/browse/ED-1586

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,6 +53,7 @@
     "@types/crypto-js": "^4.1.1",
     "@types/lodash": "^4.14.191",
     "@types/marked": "^4.0.1",
+    "@types/nprogress": "^0.2.3",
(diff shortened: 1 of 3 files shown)


Kinplemelon:
Now there are still some files that use JS, and if we need to convert all of them to TS we may need a new task

ysfscream (approved):

ysfscream:
> Now there are still some files that use JS, and if we need to convert all of them to TS we may need a new task

GOT IT. We convert important files like HTTP ones first. Others will be handled later.

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-6849)

PR #2975, opened 2024-09-25
fix(banned): adapt infinity

Kinplemelon:
https://emqx.atlassian.net/browse/ED-1616


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-4302)

PR #2986, opened 2024-09-27
chore: add a default(true) value for enable when creating rule

zhongwencool:
Every time a new rule is created, it is enabled by default, but there is no 'enable' field. So, let's add a default value. :)
> This change has no impact on users; other fields (description, sql) have default values when creating a new rule, but this one does not, which feels a bit Inconvenient and inconsistent.


Kinplemelon (approved):

ysfscream (approved):

ysfscream:
Thanks 文哥 @zhongwencool 👍

PR #2996, opened 2024-10-09
refactor(connector): hide `ssl` for IoTDB thrift

Kinplemelon:
In version 581, the IoTDB thrift driver needs to hide the ssl configuration, the dashboard also removes the ssl configuration when submitting for insurance. After confirming, we can also submit the default ssl configuration, just make sure `ssl.enable` is `false`


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-4405)

0721Betty (approved):

PR #3013, opened 2024-10-18
fix(authn): authn detail page mechanism shows value

Kinplemelon:
https://emqx.atlassian.net/browse/ED-1640


ysfscream (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-4539)

PR #3029, opened 2024-10-30
fix(clients): label for filtering options

Kinplemelon:
https://emqx.atlassian.net/browse/ED-1655


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-4826)

ysfscream (approved):

PR #3035, opened 2024-11-04
refactor(global): remove spaces before and after form values

Kinplemelon:
https://emqx.atlassian.net/browse/ED-1654

https://emqx.atlassian.net/browse/ED-1045


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-4883)

PR #3037, opened 2024-11-05
fix(rule): highlighting of rule sql

Kinplemelon:
https://emqx.atlassian.net/browse/ED-487


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-4889)

PR #3059, opened 2024-11-18
fix(listener): repeated configs in custom configs

Kinplemelon:
https://emqx.atlassian.net/browse/ED-1676


ysfscream (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-5101)

PR #3071, opened 2024-12-02
fix(mqtt & listener): update desc from emqx

Kinplemelon:
https://emqx.atlassian.net/browse/ED-1692


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-5234)

ysfscream (approved):

PR #3073, opened 2024-12-03
ci: use GitHub App to create PRs in emqx/emqx

id:
This is to get rid of PAT, and use ephemeral github app tokens instead.


ysfscream (approved):
LGTM

PR #3075, opened 2024-12-03
fix(listener): desc for mountpoint

Kinplemelon:
https://emqx.atlassian.net/browse/ED-1694


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-5249)

ysfscream (approved):

PR #3079, opened 2024-12-05
fix(schema): validate schema after generate from JSON

Kinplemelon:
https://emqx.atlassian.net/browse/ED-1697


ysfscream (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-5358)

PR #3081, opened 2024-12-06
fix(rule): support jump to action detail page

Kinplemelon:
https://emqx.atlassian.net/browse/ED-1680


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-5353)

PR #3109, opened 2024-12-18
chore(utils): update shared-ui-utils to 0.0.16

ysfscream:
Fixed IoTDB batch template issue: https://emqx.atlassian.net/browse/ED-1713


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-5570)

PR #3110, opened 2024-12-18
fix(action): remove key if value is empty

Kinplemelon:


ysfscream (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-5569)

PR #3111, opened 2024-12-18
fix(action): default value of boolean enums

Kinplemelon:


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-5571)

PR #3115, opened 2024-12-19
fix(flow): add tablestore node

ysfscream:
https://emqx.atlassian.net/browse/ED-1719


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-5572)

PR #3116, opened 2024-12-19
fix(schema): optimize input with placeholder select

Kinplemelon:


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-5573)

PR #3117, opened 2024-12-19
fix(connector & action): issue when saving as copy; simple oneof

Kinplemelon:

https://emqx.atlassian.net/browse/ED-1716


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-5498)

PR #3123, opened 2024-12-23
fix(schema): do not show boolean options if there are other type

Kinplemelon:
https://emqx.atlassian.net/browse/ED-1726


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-5548)

ysfscream (approved):

PR #3125, opened 2024-12-23
fix(monitoring): update large heap desc

ysfscream:
https://emqx.atlassian.net/browse/ED-1725


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-5576)

PR #3131, opened 2024-12-24
fix(nodes): fix cluster name too long issue

ysfscream:
https://emqx.atlassian.net/browse/ED-1722


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-5581)

Kinplemelon:
Can we use https://github.com/emqx/emqx-dashboard5/blob/ysf/1.8.5/src/components/CommonOverflowTooltip.vue to solve this problem?

ysfscream:
@Kinplemelon Yes. I can! I will try replace it.

ysfscream:
@Kinplemelon Updated

PR #3135, opened 2024-12-25
fix(connector): support static client ids for mqtt connector

ysfscream:
https://emqx.atlassian.net/browse/ED-1723


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-5903)

PR #3176, opened 2025-01-21
fix(webhook): toggle connector enable state when toggling webhook ena…

Kinplemelon:
…ble state

https://emqx.atlassian.net/browse/ED-1746


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-6147)

PR #3177, opened 2025-01-21
Summary API

Kinplemelon:
It's not quite finished, but we can merge it first 😬


src/i18n/Base.js
```diff
@@ -643,4 +643,8 @@ export default {
     zh: '，',
     en: ', ',
   },
+  lastModified: {
+    zh: '最后修改',
```
ysfscream, line 647: 更新时间？

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-6144)

PR #3194, opened 2025-02-10
fix(tong auth guidance): correct config

Kinplemelon:


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-6065)

ysfscream (approved):

PR #3195, opened 2025-02-11
fix(listener): add exit confirmation

Kinplemelon:
https://emqx.atlassian.net/browse/ED-1763

https://emqx.atlassian.net/browse/ED-1759


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-6152)

PR #3200, opened 2025-02-12
fix(flow): MQTT action and source have the same name

Kinplemelon:
https://emqx.atlassian.net/browse/ED-1758


ysfscream (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-6156)

PR #3204, opened 2025-02-13
fix(action): logic to detect if it is a webhook action

Kinplemelon:
https://emqx.atlassian.net/browse/ED-1767

https://emqx.atlassian.net/browse/ED-1768


ysfscream (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-6161)

PR #3209, opened 2025-02-17
fix(action): issue of overflow tooltip

Kinplemelon:
https://emqx.atlassian.net/browse/ED-1770


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-6165)

PR #3221, opened 2025-02-19
refactor(component): replace select with input tag in array editor

Kinplemelon:
https://emqx.atlassian.net/browse/ED-1757


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-6181)

ysfscream (approved):

PR #3226, opened 2025-02-20
fix(websocket): check topic before pub and sub

Kinplemelon:
same as https://emqx.atlassian.net/browse/ED-1777


ysfscream (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-6872)

PR #3227, opened 2025-02-20
fix(msg trans): message after deleting

Kinplemelon:
https://emqx.atlassian.net/browse/ED-1780


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-6189)

ysfscream (approved):

PR #3230, opened 2025-02-21
fix(rule): reg to match event type

Kinplemelon:
https://emqx.atlassian.net/browse/ED-1778


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-6212)

ysfscream (approved):

PR #3240, opened 2025-02-28
fix(audit log): correct some text in dict

Kinplemelon:
https://emqx.atlassian.net/browse/ED-1784


ysfscream (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-6878)

PR #3249, opened 2025-03-03
chore(ci): use `ref` to detect version

Kinplemelon:


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-6885)

PR #3267, opened 2025-03-06
fix(webhook): unexpected error message

Kinplemelon:
https://emqx.atlassian.net/browse/ED-1785


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-6380)

PR #3301, opened 2025-04-17
fix(rule): can not refresh rule list

Kinplemelon:
https://emqx.atlassian.net/browse/ED-1854

https://emqx.atlassian.net/browse/ED-1839


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-7071)

ysfscream (approved):

PR #3305, opened 2025-04-18
fix(namespace): allow delete implicitly namespace

Kinplemelon:
https://emqx.atlassian.net/browse/ED-1861


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-7080)

ysfscream (approved):

PR #3306, opened 2025-04-18
fix(namespace): add word-break style for better text handling

ysfscream:
https://emqx.atlassian.net/browse/ED-1860
https://emqx.atlassian.net/browse/ED-1863


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-7099)

PR #3310, opened 2025-04-21
fix(i18n): add a space between the QoS and the number

Kinplemelon:
https://emqx.atlassian.net/browse/ED-1823


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-7109)

PR #3322, opened 2025-04-24
fix(listener): set to null if limiter value is empty

Kinplemelon:
https://emqx.atlassian.net/browse/ED-1884


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-7173)

ysfscream (approved):

PR #3331, opened 2025-05-06
fix(rule): can not update repub fallback action

Kinplemelon:
https://emqx.atlassian.net/browse/ED-1897


ysfscream (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-7454)

PR #3333, opened 2025-05-07
Sync Code From 591 to Enterprise

Kinplemelon:


shirley611:
[TestExecution](https://emqx.atlassian.net/browse/QA-7359)

PR #3362, opened 2025-05-28
fix(flow ai): comment out 'o1-pro' option in model options

Kinplemelon:
https://emqx.atlassian.net/browse/EMQX-14308


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-7665)

PR #3365, opened 2025-05-29
ci: translate issue from Chinese to English

Rory-Z:


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-7668)

PR #3368, opened 2025-05-30
fix(schema registry): http configs

Kinplemelon:
https://emqx.atlassian.net/browse/ED-1948

https://emqx.atlassian.net/browse/ED-1947


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-7673)

PR #3369, opened 2025-05-30
refactor(ruleengine): update generateSQLByAI to use fetch API

ysfscream:
https://emqx.atlassian.net/browse/ED-1911


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-7672)

PR #3376, opened 2025-06-06
fix(api): update fetch URL in generateSQLByAI

ysfscream:


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-7728)

PR #3380, opened 2025-06-09
fix(gateway): correct desc for client info

Kinplemelon:
https://emqx.atlassian.net/browse/ED-1968


ysfscream (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-7737)

PR #3384, opened 2025-06-09
fix(flow): rule actions

Kinplemelon:


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-7750)

PR #3385, opened 2025-06-10
fix(schema registry): cannot create normal schema registry

Kinplemelon:
https://emqx.atlassian.net/browse/ED-1978


ysfscream (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-7751)

PR #3388, opened 2025-06-10
fix(connector): default value for `middlebox_comp_mode`

Kinplemelon:


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-7882)

PR #3399, opened 2025-06-24
fix(listener): incorrect unsaved prompt

Kinplemelon:
https://emqx.atlassian.net/browse/ED-1987


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-8040)

PR #3402, opened 2025-06-30
fix(flow): label for function parameter

Kinplemelon:
https://emqx.atlassian.net/browse/ED-1996


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-8094)

ysfscream (approved):

PR #3403, opened 2025-06-30
fix(flow): label for function parameter

Kinplemelon:


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-8226)

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-8226)

PR #3410, opened 2025-07-11
refactor(flow): add back some gemini models

Kinplemelon:


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-8291)

PR #3421, opened 2025-08-11
fix(connector & action): use textarea for `private_key`

Kinplemelon:
https://emqx.atlassian.net/browse/ED-2080


ysfscream (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-8630)

PR #3426, opened 2025-08-18
refactor(connector): adapt datalayers

Kinplemelon:


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-8654)

PR #3427, opened 2025-08-18
fix(action): check connector value

Kinplemelon:


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-8656)

PR #3429, opened 2025-08-19
fix(connector): can not edit `driver_type` when creating

Kinplemelon:


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-8663)

PR #3432, opened 2025-08-20
fix(flow): correct ai type when requesting models

Kinplemelon:


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-8704)

PR #3440, opened 2025-08-29
fix(gateway): no response when adding authn in OCPP

Kinplemelon:


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-9214)

PR #3446, opened 2025-09-09
fix(connector): two issue in mqtt connector

Kinplemelon:
https://emqx.atlassian.net/browse/ED-2137


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-9364)

ysfscream (approved):

PR #3462, opened 2025-09-19
fix(message queue): desc for key expression

Kinplemelon:
https://emqx.atlassian.net/browse/ED-2165


ysfscream (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-9683)

PR #3467, opened 2025-09-24
fix(log trace): sometimes logs may not display fully

Kinplemelon:
https://emqx.atlassian.net/browse/ED-2170


ysfscream (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-9735)

PR #3473, opened 2025-09-28
style: update table header text color to #1D1D1D

ysfscream:


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-10644)

PR #3475, opened 2025-10-09
style: update left bar collapse button icon and hover style

ysfscream:


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-10645)

PR #3476, opened 2025-10-09
style: update layout dimensions and menu styling

ysfscream:
- Update header height from 48px to 49px
- Update menu collapse width from 71px to 65px
- Update menu text colors (normal: #343741, active: #1d1d1d)
- Update fold button icon size to 20px with hover styles
- Adjust collapsed menu icon positioning
- Update footer menu item padding
- Fix user avatar size in nav header

### DIFF
--- a/src/style/normalize.scss
+++ b/src/style/normalize.scss
@@ -47,7 +47,7 @@
     'bg-color': var(--color-bg-content),
   ),
   $header: (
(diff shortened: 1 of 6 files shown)


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-10646)

PR #3477, opened 2025-10-09
refactor(i18n): simplify overview tab label to avoid repetition

ysfscream:
Change the overview tab label from "Cluster Overview" to "Overview" to prevent redundancy in the UI where "Cluster Overview" appeared three times (breadcrumb twice and tab once).


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-10647)

PR #3481, opened 2025-10-11
fix: resolve TypeScript error and update login page styling

ysfscream:
- Add LdapLoginCredentials interface to properly type LDAP login credentials
- Fix TypeScript error: Property 'username' does not exist on type 'SsoLogin'
- Remove unnecessary 'any' type and FIXME comment in useSSO.ts
- Set login page left sidebar to use fixed dark background color (#212429)


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-10650)

PR #3493, opened 2025-10-22
Sync code from enterprise-1.9.x

Kinplemelon:


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-10189)

PR #3525, opened 2025-11-07
fix(flow): cannot save after switching editing mode

Kinplemelon:
https://emqx.atlassian.net/browse/ED-2212


ysfscream (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-10668)

PR #3530, opened 2025-11-10
fix(object array editor): fixed the removed data index

Kinplemelon:
https://emqx.atlassian.net/browse/ED-2208


ysfscream (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-11068)

PR #3538, opened 2025-11-12
fix(action): influxdb config in detail page

Kinplemelon:


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-10505)

ysfscream (approved):

PR #3539, opened 2025-11-13
fix(schema registry): issue with changing type when editing schema registry

Kinplemelon:
https://emqx.atlassian.net/browse/ED-2222

https://emqx.atlassian.net/browse/ED-2223


ysfscream (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-10514)

PR #3541, opened 2025-11-13
fix(action): error `tags` in opents

Kinplemelon:
https://emqx.atlassian.net/browse/ED-2227


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-10513)

ysfscream (approved):

PR #3543, opened 2025-11-14
refactor(action): adapt new configs in azure blob

Kinplemelon:
https://emqx.atlassian.net/browse/ED-2225


ysfscream (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-10675)

PR #3556, opened 2025-11-19
style: update message-queue menu icon to Layers

ysfscream:


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-10728)

PR #3559, opened 2025-11-20
fix(listener): can not update tls method

Kinplemelon:
https://emqx.atlassian.net/browse/ED-2239


ysfscream (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-10730)

PR #3567, opened 2025-11-25
fix(authz): issue when adding acl rule

Kinplemelon:
https://emqx.atlassian.net/browse/ED-2258


ysfscream (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-10799)

PR #3568, opened 2025-11-27
refactor(overview): improve nodes graph layout and zoom handling

ysfscream:
- Extract FLOW_DEFAULT_ZOOM constant for consistent zoom configuration
- Adjust core node height calculation for better spacing
- Add safe radius limit to prevent nodes from overflowing viewport
- Replace hardcoded zoom values with constant

https://emqx.atlassian.net/browse/ED-2261


Kinplemelon (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-10914)

PR #3594, opened 2025-12-19
fix(rule): style issue

Kinplemelon:
https://emqx.atlassian.net/browse/ED-2313


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-11105)

PR #3611, opened 2026-01-14
fix(rule): issues when parsing foreach

Kinplemelon:
https://emqx.atlassian.net/browse/ED-2334


ysfscream (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-12101)

PR #3616, opened 2026-01-15
fix(listener): no session_tickets for QUIC

Kinplemelon:


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-11541)

PR #3618, opened 2026-01-15
fix(listener): remove reuse_sessions from unexposed options

Kinplemelon:


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-12106)

PR #3625, opened 2026-01-20
chore(dependencies): update @emqx/shared-ui-i18n to version 0.0.63

Kinplemelon:


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-11710)

PR #3629, opened 2026-01-26
fix(gateway): show message-related metrics in JT808

Kinplemelon:
https://emqx.atlassian.net/browse/ED-2359
https://emqx.atlassian.net/browse/ED-2366
https://emqx.atlassian.net/browse/ED-2364


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-11799)

PR #3633, opened 2026-01-30
fix(TLSEnableConfig): update form item property to reuse_sessions

Kinplemelon:
thanks for @0721Betty


ysfscream (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-12118)

PR #3640, opened 2026-02-13
fix(source): node reconnection

Kinplemelon:


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-12201)

ysfscream (approved):

PR #3641, opened 2026-02-24
fix(streams & queues): handle `auto` for enable

Kinplemelon:
https://emqx.atlassian.net/browse/ED-2396


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-12240)

ysfscream (approved):

PR #3643, opened 2026-02-25
feat(plugin): init ui tab

Kinplemelon:


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-12254)

PR #3648, opened 2026-03-02
fix(i18n): update login locked message for clarity

Kinplemelon:


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-12538)

PR #3677, opened 2026-03-30
refactor(connector): add label and description for new audience confi…

Kinplemelon:
…g in GCP connector

use local label and description first


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-13055)

PR #3693, opened 2026-04-24
fix(user): restrict SSO users from disabling MFA

Kinplemelon:


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-13769)

ysfscream (approved):

PR #3698, opened 2026-05-09
fix(plugin): correctly display error messages

Kinplemelon:
https://emqx.atlassian.net/browse/ED-2468


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-15613)

PR #3700, opened 2026-05-11
feat(user): adapt user scopes

Kinplemelon:


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-14059)

PR #3714, opened 2026-05-28
fix(scopes): handle legacy unset dashboard scopes

Kinplemelon:
Normalize dashboard user and API key scopes when the backend returns the legacy "unset" sentinel. Display unset scopes as the default "All" state with a warning tooltip, and avoid submitting unset or empty scope lists back to the API so the backend can apply role defaults.

Also update scope response types and clarify the API key scope placeholder copy.


ysfscream (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-14662)

PR #3715, opened 2026-05-29
fix(scopes): polish legacy unset scopes handling

Kinplemelon:


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-14663)

PR #3716, opened 2026-05-31
fix(api key): set scopes when changing publisher to admin

Kinplemelon:


ysfscream (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-14664)

PR #3722, opened 2026-06-02
feat(cluster-link): add desc form dispatch_strategy in resource_opts

Kinplemelon:


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-14699)

PR #3724, opened 2026-06-03
fix(api key): handle role switch

Kinplemelon:


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-14719)

PR #3737, opened 2026-06-17
fix(connector): allow file in gcp connector

Kinplemelon:


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-15806)

PR #3751, opened 2026-07-01
fix(authn): correct descriptions for `on_missing_jwt` configuration

Kinplemelon:


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-16021)

PR #3758, opened 2026-07-09
feat(topic metrics 2): new topic metrics

Kinplemelon:
fixes https://github.com/emqx/emqx-dev-team-tasks/issues/262

### DIFF
--- a/src/i18n/Tools.ts
+++ b/src/i18n/Tools.ts
@@ -139,6 +139,46 @@ export default {
     zh: '添加主题',
     en: 'Add Topic',
   },
+  topicMetricsCollection: {
+    zh: '主题监控集合',
+    en: 'topic metrics collection',
+  },
+  createCollection: {
+    zh: '创建集合',
+    en: 'Create Collection',
+  },
+  collectionName: {
+    zh: '集合名称',
+    en: 'Collection Name',
+  },
+  collectionNameError: {
+    zh: '集合名称仅支持字母、数字、短横线和下划线，最长 64 个字符',
+    en: 'The collection name can contain only letters, digits, hyphens, and underscores, up to 64 characters.',
+  },
+  topicFilter: {
+    zh: '主题过滤器',
+    en: 'Topic Filter',
+  },
+  bytesIn: {
+    zh: '流入字节',
+    en: 'Bytes In',
+  },
+  bytesOut: {
+    zh: '流出字节',
+    en: 'Bytes Out',
+  },
+  resetCollectionTip: {
+    zh: '确定要重置集合 {name} 的指标吗？',
+    en: 'Are you sure you want to reset metrics for collection {name}?',
+  },
+  deleteCollectionTip: {
+    zh: '确定要删除集合 {name} 吗？',
+    en: 'Are you sure you want to delete collection {name}?',
+  },
+  topicFilterMountpointTip: {
+    zh: '当前用户属于命名空间 {namespace}。主题过滤器匹配的是应用 mountpoint 后的 topic，例如客户端发布 t/x 时，需要填写 {namespace}/t/x 或 {namespace}/#。',
+    en: 'The current user belongs to namespace {namespace}. The topic filter matches the topic after the mountpoint is applied. For example, when a client publishes t/x, use {namespace}/t/x or {namespace}/#.',
+  },
   msgIn: {
     en: 'Published Messages',
     zh: '消息发布',
--- a/src/i18n/Tools.ts
+++ b/src/i18n/Tools.ts
@@ -176,8 +176,8 @@ export default {
     en: 'Are you sure you want to delete collection {name}?',
   },
   topicFilterMountpointTip: {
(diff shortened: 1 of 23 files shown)


ysfscream:
Hi @Kinplemelon 👋 The PR description is empty. Could you add a few screenshots to help the review? The list page with data, the Create Collection dialog, and an expanded row would cover most of it. Thanks!

src/views/Diagnose/TopicMetricsCollections.vue
```diff
@@ -0,0 +1,395 @@
...
+const tableExpandRowKeys = computed(() =>
+  collections.value.filter(({ _expand }) => _expand).map(getRowKey),
+)
+
+const isOpNsDisabled = (row: TopicMetricCollectionRow) => isOpNsResourceDisabled(row)
+const getMetric = ({ metrics }: TopicMetricCollectionRow, key: MetricKey) => metrics?.[key] ?? 0
+const formatCount = (value?: number) => Number(value ?? 0).toLocaleString()
+const formatBytes = (value?: number) => transMemorySizeNumToStr(Number(value ?? 0), 2)
+
+const loadCollections = async () => {
```
ysfscream, line 263: After create/delete, `loadCollections()` replaces the rows unsorted, but the Element Plus header keeps its sort arrow (`sortable="custom"`) — so the header and data disagree. Consider re-applying the current sort at the end of the reload, or calling `tableRef?.clearSort()`.

src/views/Diagnose/TopicMetricsCollections.vue
```diff
@@ -0,0 +1,395 @@
...
+  !namespace || namespace === GLOBAL_NAMESPACE ? t('BasicConfig.global') : namespace
+
+const getRowKey = ({ namespace, name }: TopicMetricCollectionRow) =>
+  `${namespace || GLOBAL_NAMESPACE}::${name}`
+
+const tableExpandRowKeys = computed(() =>
+  collections.value.filter(({ _expand }) => _expand).map(getRowKey),
+)
+
+const isOpNsDisabled = (row: TopicMetricCollectionRow) => isOpNsResourceDisabled(row)
```
ysfscream, line 233: `isOpNsDisabled` only forwards to `isOpNsResourceDisabled` with no added logic. Could use `isOpNsResourceDisabled` directly in the template and drop the wrapper.

src/views/Diagnose/TopicMetricsCollections.vue
```diff
@@ -0,0 +1,395 @@
...
+              <div :class="isNamespaceUser ? 'col-start-2' : 'col-start-3'">
+                <p class="m-0 mb-2 text-[var(--color-text-secondary)]">{{ tl('bytesOut') }}</p>
+                <p class="m-0 text-base font-normal text-[var(--color-text-primary)]">
+                  {{ formatBytes(getMetric(row, 'bytes.out')) }}
+                </p>
+              </div>
+              <div :class="isNamespaceUser ? 'col-start-4' : 'col-start-5'">
+                <p class="m-0 mb-2 text-[var(--color-text-secondary)]">{{ t('Base.createdAt') }}</p>
+                <p class="m-0 text-base font-normal text-[var(--color-text-primary)]">
+                  {{ row.create_time ? dateFormat(row.create_time, '') : '-' }}
```
ysfscream: This can be simplified to `{{ dateFormat(row.create_time, -) }}` — `dateFormat` already falls back for null/undefined/invalid values, and it also fixes the case where a non-empty but invalid date currently renders as an empty string.

src/views/Diagnose/TopicMetricsCollections.vue
```diff
@@ -0,0 +1,395 @@
...
+
+type MetricKey = keyof TopicMetricCollectionMetrics
+type TopicMetricCollectionRow = TopicMetricCollection & {
+  _expand?: boolean
+  _loading?: boolean
+}
+
+const { t, tl } = useI18nTl('Tools')
+const store = useStore()
+const isNamespaceUser = computed(() => store.getters.isNamespaceUser)
```
ysfscream, line 213: `useNsResource()` below already derives `isNamespaceUser` from the same store getter, so this line duplicates it. Minor — just a chance to keep a single source.

src/i18n/Tools.ts
```diff
@@ -139,6 +139,46 @@ export default {
     zh: '添加主题',
     en: 'Add Topic',
   },
+  topicMetricsCollection: {
```
ysfscream, line 142: The `en` value is lowercase here while the sibling keys use Title Case (`Bytes In`, `Collection Name`). Since it is shown as a label, suggest `Topic Metrics Collection`.

ysfscream (commented):
Nice work — the feature is clean and consistent with the existing `TopicMetrics.vue`. A few minor, non-blocking notes inline: one sorting edge case worth a look, the rest are small simplifications.

ysfscream:
Could you also link the related task issue in the PR description (e.g. `Closes #xxx` or a reference)? It makes tracking and cross-linking easier. Thanks!

ysfscream (approved):

PR #3968, opened 2026-09-07
feat(auth): support SCRAM dashboard login

Kinplemelon:


src/common/scramCryptoJs.ts
```diff
@@ -0,0 +1,35 @@
+import type { ScramProof, ScramProofInput } from '@/types/scram'
+import CryptoJS from 'crypto-js'
```
ysfscream: Pulls the whole library (~25 KB gz, all ciphers) into the worker chunk — which loads precisely in weaker environments (plain-HTTP deployments). Repo convention (`src/common/auth.ts`) is submodule imports: `crypto-js/core` + `sha256` + `hmac-sha256` + `pbkdf2` + `enc-base64` (~5 KB gz).

src/common/scramCore.ts
```diff
@@ -0,0 +1,106 @@
...
+  }
+  return btoa(binary)
+}
+
+const isValidNonce = (value: string, maxLength: number) =>
+  value.length >= MIN_NONCE_LENGTH && value.length <= maxLength && NONCE_PATTERN.test(value)
+
+type CompleteScramChallenge = Required<ScramChallenge>
+
+const validateChallenge: (
```
ysfscream, line 44: ~40 lines of per-field checks whose distinct messages all funnel into one generic toast; some check our own output (the client nonce is created here as exactly 32 base64url chars, yet range-checked 20–128) or restate limits the generated schema already documents. A single type guard + one throw does the same job:

```ts
const isCompleteChallenge = (c: ScramChallenge): c is Required<ScramChallenge> =>
  c.mechanism === SCRAM_SHA_256 &&
  typeof c.server_nonce === 'string' && c.server_nonce.length > 0 && c.server_nonce.length <= MAX_NONCE_LENGTH &&
  Number.isInteger(c.iterations) && (c.iterations as number) >= MIN_ITERATIONS && (c.iterations as number) <= MAX_ITERATIONS &&
  typeof c.salt === 'string' && base64ToBytes(c.salt).length === SCRAM_SALT_BYTES
```

Non-blocking.

src/common/scramCore.ts
```diff
@@ -0,0 +1,106 @@
...
+  const combinedNonce = clientNonce + challenge.server_nonce
+  const escapedUsername = escapeScramUsername(username)
+  const authMessage =
+    `n=${escapedUsername},r=${clientNonce}` +
+    `,r=${combinedNonce},s=${challenge.salt},i=${challenge.iterations}` +
+    `,c=biws,r=${combinedNonce}`
+  return {
+    password,
+    salt: base64ToBytes(challenge.salt),
+    saltBase64: challenge.salt,
```
ysfscream: `salt` and `saltBase64` are one datum in two representations, and each consumer must know which to pick. Keep only `salt: Uint8Array` — CryptoJS takes it directly via `CryptoJS.lib.WordArray.create(prepared.salt)` (needs `import 'crypto-js/lib-typedarrays'`).

src/common/scram.ts
```diff
@@ -0,0 +1,150 @@
+import type { ScramProof, ScramProofInput, ScramWorkerResponse } from '@/types/scram'
+import { base64ToBytes, bytesToBase64, prepareScramProof, ScramLoginError } from './scramCore'
+
+export { base64ToBytes, bytesToBase64, prepareScramProof, ScramLoginError } from './scramCore'
```
ysfscream: Only `ScramLoginError` is consumed through this re-export, and `scramCryptoJs.ts` imports `./scramCore` directly — two import styles for the same module. Suggest deleting this line and importing `ScramLoginError` from `@/common/scramCore` at the two call sites.

src/types/scram.ts
```diff
@@ -0,0 +1,21 @@
+import type { ScramChallenge } from './typeAlias'
```
ysfscream, line 1: This 21-line module adds a third hop to every SCRAM import chain. The three interfaces + const could live in `scramCore.ts` (already the shared dependency of both crypto backends) — 5 new files become 4 without affecting the worker chunk split. Non-blocking.

src/api/scram.ts
```diff
@@ -0,0 +1,53 @@
...
+  verifyScramServerSignature,
+} from '@/common/scram'
+import type {
+  ScramChallenge,
+  ScramChallengeRequest,
+  ScramLoginResponse,
+  ScramVerifyRequest,
+} from '@/types/typeAlias'
+
+interface LoginCredentials {
```
ysfscream: Third copy of this shape — `login()` in `api/common.ts` and `queryLogin` in Login.vue both inline it. Worth one shared type so they can't drift. Also `requestScramChallenge`/`requestScramVerification` are only used inside `scramLogin`; they can stay unexported.

src/api/scram.ts
```diff
@@ -0,0 +1,53 @@
...
+  username: string
+  mfa_token?: string
+}
+
+export const requestScramChallenge = (
+  username: string,
+  clientNonce: string,
+): Promise<ScramChallenge> => {
+  const request: ScramChallengeRequest = { username, client_nonce: clientNonce }
+  return http.post('/login/challenge', request, { keepSpaces: true })
```
ysfscream: `keepSpaces` was carried over from the old `login()`, which needed it because the body held the raw password (the interceptor trims strings by default). Here the password never transits and the username is already `v-model.trim`'d — suggest dropping it rather than leaving readers to invent a rationale.

src/api/scram.ts
```diff
@@ -0,0 +1,53 @@
...
+}
+
+export const requestScramVerification = (
+  request: ScramVerifyRequest,
+): Promise<ScramLoginResponse> => http.post('/login/verify', request)
+
+export const scramLogin = async (credentials: LoginCredentials): Promise<ScramLoginResponse> => {
```
ysfscream: `scramLogin` is protocol orchestration (challenge → derive → verify → signature check) sitting in `src/api/`, where neighbors are thin `http.x` wrappers. Consider keeping this file to the two endpoint calls and moving the orchestration next to the crypto in `common/scram.ts`. Non-blocking.

src/api/scram.ts
```diff
@@ -0,0 +1,53 @@
...
+  const { username, password } = credentials
+  const mfaToken = credentials.mfa_token || undefined
```
ysfscream: `queryLogin` already normalizes `mfa_token || undefined` before calling — this repeats it. `...(credentials.mfa_token ? { mfa_token: credentials.mfa_token } : {})` inline (or pass through; axios drops `undefined` keys) keeps the munging in one layer.

src/views/Base/Login.vue
```diff
@@ -543,6 +544,9 @@ const queryLogin = async (user: { username: string; password: string; mfa_token?
     if (code === MFA_REQUIRED) {
       handleMFAMethod(message, username)
     }
+    if (error instanceof ScramLoginError) {
```
ysfscream: The two error families are disjoint — a `ScramLoginError` never has `error.response.data`. An early return makes that obvious and keeps a diagnostic trail:

```ts
if (error instanceof ScramLoginError) {
  console.error(error)
  ElNotification.error(t('Base.scramLoginError'))
  isSubmitting.value = false
  return Promise.reject(error)
}
```

ysfscream (commented):
Nice work — the crypto itself looks right (validation, constant-time compare, worker fallback). The inline comments are simplifications; three higher-level things to decide before merge:

1. **The protocol switch is hardcoded at build time.** All logins now go through SCRAM via an import-alias swap, but the new `dashboard.password_login` config (`both | scram_only`) and 403 `PASSWORD_LOGIN_DISABLED` have no consumers, and `login()` in `api/common.ts` becomes dead code. Either commit to unconditional SCRAM (drop the old API) or read the config and dispatch.

2. **Error codes still derive from the legacy endpoint.** `LOGIN_LOCKED` comes from `PostLogin401Code` (the old `/login` schema) and works only because the strings coincide with `PostLoginVerify401Code`. `SCRAM_CHALLENGE_INVALID` is unhandled — an expired/replayed challenge will surface as a raw backend toast.

3. **Every ScramLoginError flattens into one generic toast with nothing logged.** That includes server-signature mismatch, a possible MITM signal. At minimum `console.error(error)` in the catch.

Non-blocking: the schema regen (rocketmq_namespace, resolved_address, otel protocol, audit) and auto-imports.d.ts are unrelated churn — the repo usually lands those as standalone schema-sync commits, and splitting would shrink this diff a lot. Also `MAX_ITERATIONS = 10M` is unreachable for the pure-JS fallback within its 55s worker timeout; fine for same-vendor backends, just noting.

ysfscream (approved):
LGTM. Note: merging with check_audit_dict red — the audit dict entries for /login/challenge and /login/verify still need a follow-up commit.

